PR #1, opened 2025-07-30
APPENG-3461 Agent V1

fax4ever:
https://issues.redhat.com/browse/APPENG-3461

As you know better than I do, in programming there are countless ways to approach a problem—many of which are equally valid.
So please consider this contribution as a starting point for discussion.


backend/Containerfile
```diff
@@ -0,0 +1,18 @@
+FROM registry.access.redhat.com/ubi9/python-312:latest
+
```
mhdawson, line 2: I'd suggest that we move Container file from the backend directory into the agent-manager directory and remove the backend as a top level directory. (ie make agent-manager be at the top level for now. Then including the src directory as you have in the process/standards doc in agent-manager make sense.

backend/agent-manager/src/agent_manager/agent_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            for agent in self._config["agents"]:
+                self.create_agent(agent)
+        else:
+            logging.debug("Agents already created")
+
+    def create_agent(self, agent: dict):
+        agent_config:AgentConfig={
+            "model": agent['model'],
+            "instructions": agent['instructions'],
+            "tool_choice": agent['tool_choice'],
```
mhdawson, line 38: I think we need  "toolgroups":  as well here.

mhdawson: As an example 

```python
            "toolgroups": [
                {
                    "name": "builtin::rag/knowledge_search",
                    "args": {"vector_db_ids": ["laptop-refresh-knowledge-base"]},
                },
                "mcp::asset_database",
                "mcp::servicenow",
            ],
```

We at least need to be able to specify 1 or more mcp servers and 1 or more knowledge bases.

In our configuration file we need two types of entries one for mcp server and a different one for knowlege bases. The mcp servers will map to entries like the last two toolgroups in the example above, and each knowledge base would make to an entry like there is for "laptop-refresh-knowledge-base" in the example

mhdawson: The PRD has some more specific details of what should be in the config -https://docs.google.com/document/d/1Vgxgp77H7hW2VpUVO43d2vGybbfRpPnR3ZagSsli7mg/edit?tab=t.0#heading=h.r8703jdria1q

It's ok to add/modify those, but we should keep the two in sync (PRD/Implementation)

backend/local_llamastack_server/README.md
```diff
@@ -0,0 +1,27 @@
+# Local Llama Stack Server
```
mhdawson, line 1: I'd suggest we have support for locall testing in a directory called `local_testing` at the top level. As mentioned before a config file which has the environment variables that are set in the helm deployment and a script to set them in the enviroment would be good.

mhdawson:
In terms of integrating this into the kubernes deploy, I think it might make sense to have the agent configuration run using https://kubernetes.io/docs/concepts/workloads/pods/init-containers/


I think that would mean adding another entry to https://github.com/RHEcosystemAppEng/self-service-agent-blueprint/blob/8025543a16b9f30d2397a1fb73472a433624435e/deploy/helm/self-service-agent/templates/deployment.yaml#L36, making sure that it only runs after the wait-for-llamastack init containers completes (not sure if they are sequential by default)

At some point we might detach agent config from startup but this would be a good way to start.

backend/Containerfile
fax4ever: Sure, I can do it. Regarding this change, I would like to ask you how we can use the module. For instance, in my proposal, the agent was a package module of the root project, and UV assembles the backend to make the collaboration work. Since we're going to remove Unicorn and FastAPI as well, what should be set as the entrypoint for the Containerfile? If I deploy the agent module alone, how can I interact with it from the outside? In the meantime, I'm making the change. Thanks for the help.

backend/agent-manager/src/agent_manager/agent_manager.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self):
+        self._config = load_yaml(Path(__file__).parent / "config.yaml")
+        self._client = None
+        self._agents = []
+
+    def connect_to_llama_stack(self):
+        if self._client is None:
+            logging.debug("Connecting to LlamaStack")
+            self._client = LlamaStackClient(
+                base_url=self._config['llama_stack_url'],
```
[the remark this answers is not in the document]
fax4ever: It it a great idea... going to apply it

backend/agent-manager/src/agent_manager/agent_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            "tool_choice": agent['tool_choice'],
+            "input_shields": agent['input_shields'],
+            "output_shields": agent['output_shields'],
+            "max_infer_iters": agent['max_infer_iters'],
+        }
+        agentic_system_create_response = self._client.agents.create(
+            agent_config=agent_config
+        )
+        agent_id = agentic_system_create_response.agent_id
+        self._agents.append(Agent(agent_id, agent['name'], agent['description'], agent_config))
```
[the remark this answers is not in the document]
fax4ever: great point! we have already the state on the Llamastack side. I'm going to remove the state from our side and use the remote one.
Thanks for the idea!

backend/agent-manager/src/agent_manager/config.yaml
```diff
@@ -0,0 +1,11 @@
+llama_stack_url: http://localhost:8321
```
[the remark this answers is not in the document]
fax4ever: I agree with you. I moved out the config from the src here: https://github.com/RHEcosystemAppEng/self-service-agent-blueprint/pull/1/commits/d93409b039dfe8120dbb7c3c36603e6fe4d16930.
About having more agents yaml, I agree this provides maximum flexibility and solidity with respect of the changes. Maybe we can address it a follow up?

fax4ever:
> At some point we might detach agent config from startup but this would be a good way to start.

I think that it a great idea having the registration of the configured agents as init container.
One possibility I think  could be to produce 2 container image one specific to be run as init container, having an entrypoint that simply invokes the `AgentManager#create_agents`... for instance.

backend/agent-manager/src/agent_manager/config.yaml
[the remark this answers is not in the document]
fax4ever: Sorry I changed my mind. I'm addressing in the context of this contribution....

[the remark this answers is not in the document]
fax4ever: Addressed here: https://github.com/RHEcosystemAppEng/self-service-agent-blueprint/pull/1/commits/6f08a4feceb2f495265dfbcbb5adb7ed3e641275

agent-manager/.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
mhdawson, line 1: how does this related to `FROM registry.access.redhat.com/ubi9/python-312:latest` in the container file which seems to be python 312

agent-manager/Containerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN dnf update -y && dnf clean all
+
+RUN pip3 install --no-cache-dir uv
+
+# Copy the project into the image
+COPY . /app
+
+# Install the application dependencies.
+WORKDIR /app
+RUN uv sync --frozen --no-cache --all-packages
```
mhdawson, line 13: There does not seem to be a CMD ?

I was expecting that we could run this container and it would register the agents that were defined?

agent-manager/config/agents/agent-1.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: "Agent 1 is a helpful assistant that can answer questions and help with tasks."
+model: "llama3.2:3b"
+instructions: "You are a helpful agent"
+tool_choice: "auto"
+input_shields: []
+output_shields: []
+max_infer_iters: 10
+mcp_servers:
+  - "asset_database"
+  - "servicenow"
```
mhdawson, line 11: We don't have these yet so we can't register them yet
```suggestion
```

agent-manager/config/config.yaml
```diff
@@ -0,0 +1 @@
+timeout: 120.0
```
mhdawson, line 1: Do we still need this file? I see there is a version under local testing.

agent-manager/tests/test_agent_manager.py
```diff
@@ -0,0 +1,18 @@
+import pytest
```
mhdawson, line 1: I'm thinking for now the tests directory should likely be under the local testing directory as well?

agent-manager/config/agents/agent-1.yaml
fax4ever: Applied the suggestion, thanks!

agent-manager/script/play.py
```diff
@@ -0,0 +1,24 @@
+from agent_manager.agent_manager import AgentManager
```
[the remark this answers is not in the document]
fax4ever: sure! makes sense.... doing it!

mhdawson (approved):
Pushed a few comments to allow helm deployment to work and added code to default to the deployment model to make shared development easier.

LGTM as the initial step for agent registration.

PR #5, opened 2025-08-01
feat: add simple chat script

mhdawson:
Refs: https://issues.redhat.com/browse/APPENG-3460

You can run after helm deploy with:
pod=`oc get pods |grep self-service-agent | awk '{print $1}'` 
oc exec -it $pod -- python test/chat.py

Until MCP servers and knowledge base are available it will perform poorly

### DIFF
--- a/test/chat.py
+++ b/test/chat.py
@@ -32,7 +32,7 @@ model_id = os.environ["LLAMA_STACK_MODELS"].split(",", 1)[0]
 # agent registration is done as part of the initialization
 # Once that is in place we'll need to look up the agent_id
 prompt_file = Path(__file__).resolve().parent / "prompt.txt"
-system_prompt = open(prompt_file).read()
+system_prompt = prompt_file.read_text()
 
 agentic_system_create_response = client.agents.create(
     agent_config={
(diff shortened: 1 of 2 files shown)


test/chat.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Configuration
+model_id = os.environ["LLAMA_STACK_MODELS"].split(",", 1)[0]
+
+########################
+# for now create agent, will not be needed later after
+# agent registration is done as part of the initialization
+# Once that is in place we'll need to look up the agent_id
+prompt_file = Path(__file__).resolve().parent / "prompt.txt"
+system_prompt = open(prompt_file).read()
```
TomerG2, line 35: Maybe better to use `read_text` or `with`? so the file closing is handled
```suggestion
system_prompt = prompt_file.read_text()
```

mhdawson:
@TomerG2 I think I've addressed your comments.

TomerG2 (approved):
LGTM 👍

tchughesiv:
one-liner fyi
```sh
oc exec -it deploy/self-service-agent -- python test/chat.py
```

PR #7, opened 2025-08-06
feat: mcp dummy servers

TomerG2:


mhdawson:
@TomerG2 I was thinking that the PR would include the additions to the https://github.com/RHEcosystemAppEng/self-service-agent-blueprint/tree/main/deploy/helm/self-service-agent/templates and values.yaml so that the mcp servers would be deployed along with the rest of the application.  

I assume you planning that in an follow on PR?

mcp-servers/employee-info/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Retrieves comprehensive laptop information for a specific employee.
+
+**Parameters:**
+- `employee_id` (string): The unique identifier for the employee (e.g., 'emp001')
+
+**Returns:**
+- Employee details (name, department, email)
+- Laptop specifications (brand, model, serial number, specs)
+- Assignment and warranty information
```
mhdawson, line 23: I think the key piece of information we'll need is the age of the laptop and when the warranty experies. Replacement policies might use those to figure out if the employee is elligeable for a replacement.

TomerG2: The mock data includes: `assignment_date`, `warranty_expiry`, `warranty_status`
image

TomerG2:
> I assume you planning those in an follow on PR?
Yea, this PR is already getting pretty big, I thought best to keep it focused

mcp-servers/employee-info/README.md
mhdawson: It might be useful to add the laptop age (or maybe purchase date) as I don't think any of those three would map to that, right?

mcp-servers/employee-info/README.md
```diff
@@ -0,0 +1,135 @@
...
+## Usage
+
+### Running the Server
+
+```bash
+cd mcp-servers/employee-info/
+uv run python -m employee_info.server
+```
+
+### Testing with FastMCP Client
```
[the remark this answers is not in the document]
mhdawson: I asked because I was forwared and email from I think core engineering which recommends using the official one - https://docs.google.com/document/d/1qgshTYQsLbOxiv0AmCWVpJGYr09lZkCCz_v3-P1lFL4/edit?tab=t.0

This part in particular:

For teams building with Python, we recommend using the[ official Python SDK](https://github.com/modelcontextprotocol/python-sdk). AI Engineering intends to make downstream python wheels available for this SDK for RH engineering teams to consume into their container images.

So unless we have a good reason, I'd prefer that we be using that one.

mcp-servers/employee-info/Containerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN uv sync --frozen
+
+# Expose port for HTTP transport (if needed)
+EXPOSE 8080
+
+# Set environment variables
+ENV PYTHONPATH=/app/src
+
+# Run the MCP server
+CMD ["uv", "run", "python", "-m", "employee_info.server"]
```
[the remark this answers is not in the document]
mhdawson: I think this is similar to how guidance is not to start JavaScript applications with npm (even though it often is)

mcp-servers/employee-info/src/employee_info/data/data.py
```diff
@@ -0,0 +1,79 @@
...
+                "display": "14-inch FHD",
+            },
+        },
+        "it_contact": {
+            "name": "Mike Johnson",
+            "email": "mike.johnson@company.com",
+            "phone": "+1-555-0789",
+        },
+    },
+}
```
tchughesiv, line 79: can we add more employee data with assignment dates older than 3 years? 3 years seems to be the standard time to qualify for laptop refresh

tchughesiv (changes requested):

mcp-servers/employee-info/src/employee_info/data/data.py
tchughesiv: oh, and employee location as well... EMEA, NA, LATAM, or APAC

mcp-servers/employee-info/README.md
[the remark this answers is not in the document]
mhdawson: Given that the other kickstarts use FastAPI as do the implementations in the RH IT group, I'm ok leaving it with FastAPI for now, but I think we should dicuss as a larger team whether that makes sense longer term in the context of the guidance I shared.

mhdawson:
I like the suggestion of finishing this one before adding the next one. I'm thinking that should include the additions so that it is deployed as part of the helm deploy as that might affect the pattern as well.

After that the second one should be able to a clone/change name and functions.

deploy/helm/self-service-agent/templates/employee-info-service.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.employeeInfo.enabled }}
```
mhdawson, line 1: I think that the pattern here - https://github.com/rh-ai-kickstart/ai-architecture-charts/blob/main/mcp-servers/helm/templates/service.yaml might let us have one generic service.yaml verus having to have one for each mcp server we deploy.

@TomerG2 could you confirm with @yuvalturg and if so we should adopt that pattern to be consistent with the architecture charts patter, as well as cutting down on the files we'll need.

agent-manager/src/agent_manager/agent_manager.py
```diff
@@ -41,11 +41,29 @@ def connect_to_llama_stack(self):
         else:
             logging.debug("Already connected to LlamaStack")
 
+    def register_mcp_toolgroups(self, agent):
```
mhdawson, line 44: Based on the pattern @tchughesiv introduced mcp servers should likely be a new type of asset.

mhdawson: thinking more, it may make more sense to have each MCP server self-register with llamastack as they start up

mhdawson: Also I guess that then begs the question of how do we avoid re-registration when pods restart etc. Ideally the registration would be done when the service starts ?

mhdawson: With a bit more reading an init container may not be what we want as it sounds like it will run for every intance in the pod

agent-manager/config/agents/laptop-refresh-agent.yaml
```diff
@@ -4,14 +4,17 @@ model: ""
...
 "
-tool_choice: "none"
+tool_choice: "required"
 input_shields: []
 output_shields: []
 max_infer_iters: 10
-mcp_servers: []
+mcp_servers:
+  - name: "employee-info"
+    url: "http://self-service-agent-employee-info-mcp:8000/mcp/"
```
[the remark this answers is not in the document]
mhdawson: Ok, I guess this is for the registration of the server itself.

asset-manager/src/asset_manager/agent_manager.py
```diff
@@ -45,11 +45,29 @@ def __init__(self, config):
...
+                    self._client.toolgroups.register(
+                        toolgroup_id=toolgroup_id,
+                        provider_id="model-context-protocol",
+                        mcp_endpoint={"uri": mcp_server["url"]},
+                        args={},
+                    )
+                    logging.info(f"Successfully registered toolgroup: {toolgroup_id}")
+                except Exception as e:
+                    logging.error(f"Failed to register toolgroup {toolgroup_id}: {e}")
+
```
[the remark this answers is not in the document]
tchughesiv: sure but i don't think that's what's happening here... my bigger concern was an agent creation throwing an error here because an mcp server was already registered, when really it should probably be a warning at most.

this all becomes mute though if @TomerG2 were to pivot to mcp self-registration.

mcp-servers/employee-info/src/employee_info/server.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+from typing import Dict, Any
+from mcp.server.fastmcp import FastMCP
+
+from employee_info.data import MOCK_EMPLOYEE_DATA
+from starlette.responses import JSONResponse
+
+MCP_TRANSPORT = os.environ.get("MCP_TRANSPORT", "sse")
+MCP_PORT = int(os.environ.get("SELF_SERVICE_AGENT_EMPLOYEE_INFO_SERVICE_PORT_HTTP", "8000"))
+mcp = FastMCP("Employee Info Server", host="0.0.0.0")
```
mhdawson, line 16: We should check if this can take a port as well and use MCP_PORT to set it

mcp-servers/employee-info/src/employee_info/server.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+        employee_id: The unique identifier for the employee (e.g., '1001')
+    """
+    return _get_employee_laptop_info(employee_id)
+
+
+def main() -> None:
+    """Run the Employee Info MCP server."""
+    if MCP_TRANSPORT == "sse":
```
mhdawson, line 60: We can clean this up/move this logic as the host/port is set earlier now.

mcp-servers/employee-info/src/employee_info/server.py
TomerG2: Done

mcp-servers/employee-info/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[project]
+name = "employee-info"
+version = "0.1.0"
+description = "MCP server for employee laptop information management"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "fastmcp>=2.11.2",
```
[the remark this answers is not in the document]
TomerG2: Done

mhdawson (approved):
LGTM

PR #8, opened 2025-08-06
feat: update cli to use pre-registered agents

mhdawson:
Add base agents so agent manager registers them
Update cli to use the pre-registered agents


mhdawson:
Also updated to use `pyproject.toml` instead of `requirements.txt`

PR #20, opened 2025-08-25
APPENG-3490 - Support manual release

fax4ever:
feat: APPENG-3490 - small improvement

Refs: https://issues.redhat.com/browse/APPENG-3490

`github.head_ref` is present if we're in the context of a PR, while `github.ref_name` in the case the workflow is manually started ---> we would like to support both the cases... let's see if it works


fax4ever:
> LGTM

Thank you! I'm merging it

PR #21, opened 2025-08-25
fix: APPENG-3489 - make generator wait for welcome

mhdawson:
Make the generator wait for the welcome message.
Also made the instructions more specific so flow requires ticket number to be returned.


mhdawson:
It might also be good to be able to specify the tag as well. We might want tag with a pr name in it or something like that which we can push to test a specific pr.

fax4ever (approved):
LGMT, thanks for the contribution

PR #28, opened 2025-08-28
Local deploy with kind

fax4ever:
With few changes we're able to deploy the application on a local Kind cluster.
This could be a viable option to work locally and most important to configure a light Kubernetes cluster for the CI.


fax4ever:
> instead of creating all new commands for kube, let's just change the existing commands to use `kubectl` instead, which works against both openshift and k8s clusters. `kubectl` is what `oc` is built on.

Makes sense. Doing it!

README.md
```diff
@@ -123,4 +123,23 @@ Run `make help` to see all available commands with descriptions.
...
+
+To dispose the application:
+
+```bash
+make helm-uninstall NAMESPACE=<YOUR-NAMESPACE>
+```
+
+```bash
+kubectl delete namespace <YOUR-NAMESPACE>
+```
```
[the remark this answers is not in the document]
fax4ever: you're right... at this point we don't need it anymore

tchughesiv (approved):
lgtm

helm/Chart.lock
```diff
@@ -2,14 +2,11 @@ dependencies:
 - name: pgvector
   repository: https://rh-ai-quickstart.github.io/ai-architecture-charts
   version: 0.1.0
-- name: llm-service
```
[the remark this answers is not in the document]
fax4ever: Thanks @mhdawson. Good point! You're absolutely right. Removing the possibility of deploying its own model with RHOAI would not be a good move. What I think we can assume is that in the context of our integration test, we should limit the scope to using only externally provided models. If you agree, I will try to make the dependency parametric (basically disabling it when an external model is provided). What do you think about this approach?

fax4ever:
Draft: need to fix the workflow (doing in on Monday)

README.md
```diff
@@ -16,6 +16,7 @@ A comprehensive blueprint for building self-service AI agents with modular compo
 - [uv](https://github.com/astral-sh/uv) - Fast Python package installer
 - [Podman](https://podman.io/) or Docker - Container runtime
 - [Helm](https://helm.sh/) - Kubernetes package manager (for deployment)
+- [kubectl](https://kubernetes.io/docs/reference/kubectl/) - Kubernetes Command line tool
```
TomerG2, line 19: Shouldn't kind also be added as a dependency?

fax4ever: Do you mean adding it to the application dependencies (for example, under `dependencies` in the `pyproject.toml` files)? If so, I don’t think we need it, since Kind is an infrastructure component.

If you mean on the CI server, I just fixed it by setting `using: composite` in the GitHub Action so Docker will be available on the CI runner.

TomerG2: @fax4ever oh I meant just add it here to the readme (so it's clear for devs that they should install it)

fax4ever:
I rebased the PR

fax4ever:
Hi team! With the last push:

1. Use `make test-short-integration` from the CI workflow
2. Before to deploy the application:
  * I build all the images from the local code
  * I push all the images to the local registry
3. Now the application is built using the local registry images

PR #38, opened 2025-09-12
feat: APPENG-3715 - smaller prompts test gemini

mhdawson:
- Add in token counts so that we get a feel for cost of each conversation and the evaluations
- Add in state config file using large prompt (chat-lg-state-big.yaml) instead of small one so that we can compare against smaller prompts.
- Add option to select which config file to use when running with oc
- Test with gemini and tweak small and big prompts to get good flows
- Evaluation of big/small flows are documented in https://docs.google.com/document/d/1ct63FV7Pu0UQUPoX31qhGgdoxrzk7mqJ3LJzZrZaXhQ/edit?tab=t.0

### DIFF
--- a/evaluations/helpers/run_conversation_flow.py
+++ b/evaluations/helpers/run_conversation_flow.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python3
 
+import datetime
 import json
 import logging
 import os
 from typing import Dict, List
 
 from .openshift_chat_client import OpenShiftChatClient
+from .token_counter import get_token_stats
 
 # Configure logging
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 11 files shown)


evaluations/helpers/run_conversation_flow.py
```diff
@@ -147,3 +174,66 @@ def run_flows(
                 logger.error(f"Error processing {filename}: {e}")
 
         logger.info("Flows processing completed")
+
+        # Print token usage summary with separate app and evaluation tokens
+        print("\n=== Token Usage Summary ===")
```
TomerG2, line 179: @mhdawson seems these prints are duplicated across:
  1. evaluations/helpers/run_conversation_flow.py:178-220
  2. evaluations/generator.py (similar section)
  3. evaluations/deep_eval.py (similar section)

Maybe we could create a shared `print_token_summary()` function in evaluations/helpers/token_counter.py that accepts:
  - app_tokens (dict or None for deep_eval case)
  - Optional parameters for file saving behavior

mhdawson: @TomerG2 I'll add that to my list of things to work on in follow on PRs

PR #40, opened 2025-09-15
init service now real integration

TomerG2:


mcp-servers/snow/src/snow/servicenow/client.py
```diff
@@ -0,0 +1,166 @@
...
+
+            auth_config = AuthConfig(
+                type=AuthType.API_KEY,
+                api_key=ApiKeyConfig(
+                    api_key=api_key,
+                    header_name=os.getenv(
+                        "SERVICENOW_API_KEY_HEADER", "X-ServiceNow-API-Key"
+                    ),
+                ),
+            )
```
fax4ever, line 101: Not critical, just an idea... we could introduce some sub classes (one for each if condition) for instance --- this could help in the long term maintenance.

fax4ever (commented):
Nice job @TomerG2! 
Just a suggestion from my side.

mcp-servers/snow/src/snow/server.py
```diff
@@ -82,6 +100,58 @@ def open_laptop_refresh_ticket(
...
+        raise ValueError("Employee ID cannot be empty")
+
+    if not employee_name:
+        raise ValueError("Employee name cannot be empty")
+
+    if not business_justification:
+        raise ValueError("Business justification cannot be empty")
+
+    # Determine preferred model or use default
+    model = preferred_model or "lenovo_think_pad_p_16_gen_2"
```
mhdawson, line 137: I don't think we want a default so I'd remove the "or" I know it always said optional in the description but I don't think it actually was option in the earlier versions. We should probably also update the description to say it is required.

mcp-servers/snow/src/snow/servicenow/client.py
mhdawson: One questions on that front is how much of this code is a direct copy of the original external repo. If its almost a 100% copy we might want to keep them consistent.

mhdawson (approved):
LGTM with a few suggestions/comments that can be addressed in the PR or a follow on PR.

mcp-servers/snow/src/snow/server.py
TomerG2: Done 💯

helm/values.yaml
```diff
@@ -139,6 +139,12 @@ mcp-servers:
       imageRepository: quay.io/ecosystem-appeng/self-service-agent-snow-mcp
       uri: http://self-service-agent-snow:8000/sse/
       imageTag: 0.0.2
+      env:
+        SERVICENOW_INSTANCE_URL: "https://dev295439.service-now.com/"
```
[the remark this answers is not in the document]
mhdawson: Lets keep the url out of the main Github repo

fax4ever (approved):
LGTM, thanks for the contribution!

PR #41, opened 2025-09-15
fix: APPENG-3715 - smaller prompts fix token counts

mhdawson:
- fix the counting of app tokens. Still approximate as claude tells me the Llama stack responses API implementation does not return any usage information.


fax4ever (approved):
LGTM. 
(I guess that using the tokenizer from the LLM would be too expensive.)

PR #48, opened 2025-09-17
fix: APPENG-3715 - refector chat-lg-state.py

mhdawson:
Refactor script using responses API and Langraph
into the existing Self Service agent components

Result is shared based class for session manager then one implementation when using LLama stack agents and one when for using the Responses API/Langgraph with a common Chat.py which has the option to use either session manager.

Calling chat.py uses the original session manager, challing chat-responses.py which is a simple wrapper for chat.py uses the responses session manager.


session-manager/src/session_manager/session_manager_base.py
```diff
@@ -0,0 +1,336 @@
...
+        self.agents = agent_manager.agents()
+        self.user_sessions = {}
+
+        # Optional session metadata persistence
+        self.metadata_file = None  # Subclasses can set this to enable persistence
+        self.session_metadata = {}
+
+    def _generate_session_name(self, user_id: str, agent_name: str = None) -> str:
+        """Generate a unique session name"""
+        unique_id = str(uuid.uuid4())[:8]  # First 8 characters of UUID
```
fax4ever, line 34: is it possible to have more sessions from the same user?

session-manager/src/session_manager/session_manager_responses.py
```diff
@@ -0,0 +1,309 @@
...
+            session_name: Optional session name for resuming existing conversations
+        """
+        super().__init__(agent_manager)
+
+        # Default checkpoint database path in /tmp
+        self.checkpoint_db_path = Path("/tmp/conversation_checkpoints.db")
+        self.resume_session_name = session_name
+
+        # Enable session metadata persistence
+        self.metadata_file = "/tmp/session_metadata.json"
```
fax4ever, line 47: can we address multiple sessions?

fax4ever (commented):
Very nice cleanup! Just a couple of questions (these are more like my doubts, by the way).

session-manager/src/session_manager/session_manager_base.py
mhdawson: This would depend on the integration. For the command like cli, there can be several concurrent sessions and they can each be the same user. For slack integration I think we can only have one session at a time for a given user as all chats from that user will be mapped to the existing session if it exists.

PR #50, opened 2025-09-18
Support emails

mhdawson:


mhdawson:
A few notes:

1) This does not affect the non-responses flow as it has fallbacks so that it can handle both flow types
2) This PR is adding a second e2e test ro test the responses flow as an interim measure until we switch over to it

PR #52, opened 2025-09-23
feat: APPENG-3775 - observability

fax4ever:
feat: APPENG-3775

Refs: https://issues.redhat.com/browse/APPENG-3775

To address the Llama Stack server tracing, we need also to integrate: https://github.com/rh-ai-quickstart/ai-architecture-charts/pull/84


mhdawson:
You can ignore the failure of the responses flow, I missed adding the change to Makefile to my PR, I'll fix that up this morning

helm/cluster-observability-operator/README.md
```diff
@@ -0,0 +1,171 @@
...
+| `subscription.channel` | Operator subscription channel | `stable` |
+| `subscription.packageName` | Operator package name | `cluster-observability-operator` |
+| `subscription.source` | Catalog source | `redhat-operators` |
+| `subscription.installPlanApproval` | Install plan approval mode | `Automatic` |
+| `subscription.startingCSV` | Starting CSV version | `cluster-observability-operator.v1.2.0` |
+| `operatorGroup.targetNamespaces` | Target namespaces for operator scope | `[]` (cluster-wide) |
+
+### OperatorGroup Configuration
+
+The Cluster Observability Operator is configured for cluster-wide deployment (AllNamespaces install mode) to enable comprehensive cluster monitoring and observability. The OperatorGroup uses an empty `targetNamespaces` array to achieve this scope while keeping the operator installed in its dedicated namespace.
```
mhdawson, line 57: We should probably check with Mikhail to make sure he does not have any concerns about cluster wide installation.

yuvalturg:
> My main comment is that I'm thinking it should be in
> 
> helm/charts/cluster-observability-operator/
> 
> as we are hoping this becomes part of the architecture charts once it is ready and that is more consistent with how it will be used/deployed. The only question would if that would conflict with how the existing dependent charts are updated.
> 
> @yuvalturg what's your thought on that?

@mhdawson I'd search the helm chart version of this operator and add it as a dependency to this chart.  If it doesn't exist, then yes, moving it to ai-architecture-charts is a good idea, but perhaps generalize it somehow to have the ability to install any operator from the operator hub as part of the chart.

fax4ever:
> My main comment is that I'm thinking it should be in
> 
> helm/charts/cluster-observability-operator/
> 
> as we are hoping this becomes part of the architecture charts once it is ready and that is more consistent with how it will be used/deployed. The only question would if that would conflict with how the existing dependent charts are updated.
> 
> @yuvalturg what's your thought on that?

Sounds a great idea @mhdawson! I moved it to helm/charts/cluster-observability-operator/.

fax4ever:
It seems that with the directory move, we introduced a regression:

```
ensure CRDs are installed first, resource mapping not found for name: "cluster-observability-operator" namespace: "openshift-cluster-observability-operator" from "": no matches for kind "Subscription" in version "operators.coreos.com/v1alpha1"
```

I'm testing on my fork...

fax4ever:
In the last commit I tried to place the chart under `extra-charts` in place of `charts`

fax4ever:
rebased!

mhdawson:
@fax4ever in terms of the location can you remind me of the reason for using extra-charts as the location?  Having put the mcp-servers into our repo as a work around for the issue we were seeing I'm wondering why we can't do the same for this one as well:

https://github.com/RHEcosystemAppEng/self-service-agent-blueprint/blob/54329bd31709cbc7ba7a872132168978d2aa422b/helm/Chart.yaml#L22

mhdawson (approved):
LGTM

PR #54, opened 2025-09-24
chore: fix up responses e2e testing

mhdawson:


fax4ever:
Now the error seems to be on the evalutation step.
In particular running: `make test-short-resp-integration`:

🔥 LLM EVALUATION FAILURES:
   • Total LLM failures: 1
   ❌ success-flow-1.json: Evaluation error: Evaluation LLM outputted an invalid JSON. Please use a better evaluation model.

mhdawson:
@fax4ever believe that was a once off, restarted and it passed.  There is some flakiness in both the regular and responses integration tests

fax4ever:
Thank you @mhdawson. I've merged the changes.
If the test flakiness is related to the evaluation criteria, we could make the CI checks more lenient. We can address it in a follow-up PR if it is the case.

PR #56, opened 2025-09-24
WIP: Request Manager

tchughesiv:


session-manager/src/session_manager/session_manager_responses.py
```diff
@@ -100,15 +100,13 @@ def _get_agent_for_routing(self, agent_name: str):
         return self.agent_manager.get_agent(agent_name)
 
     def _create_session_for_agent(
```
mhdawson, line 102: Was this change intended? Just wondering since I don't see any other changes in the session manager.

tchughesiv: that said, let me revert this change for now, with the understanding it will fail linter. i want to see if it fixes the e2e test that's failing

PR #59, opened 2025-09-29
fix: fallback to old mcp deployment files

mhdawson:
Changes in recent chart are not compatible with our deployments. For now fallback to previous version by adding files to our repo


tchughesiv (approved):
lgtm

PR #68, opened 2025-10-03
feat: tweak lg small prompt to work with scout

mhdawson:
Found some small tweaks that allowed the small prompt to work
with llama scout. Adding "Think step by step and use no tools" in
specific steps worked.

Added some additional options so we can limit if tools are
included in the request or not.

However, the changes seemed to break flow with 70b, so adding as a separate lg-prompt config

Signed-off-by: Michael Dawson <midawson@redhat.com>


mhdawson:
This also currently includes a partial revert of the tracing PR (removing the call in the session manager) since it seems it was imcomplete and caused crashes.

PR #72, opened 2025-10-06
fix: fail workflow if at least one eval fails

fax4ever:
if at least one eval fails then all the workflow will fail


mhdawson (approved):
LGTM

PR #73, opened 2025-10-06
fix: eval framework - reset each conversation

mhdawson:
- update eval framework to add new --reset-conversation option to reset the session at the start of each conversation. Need to run evaluations with the request manager in place

- tweak request manager so reset works correctly from eval test framework


tchughesiv (approved):
lgtm

PR #76, opened 2025-10-07
feat: Add tokens capability to request manager

tchughesiv:
Imports comprehensive token counting and tracking capability from asset-manager. Includes **tokens** command for usage statistics. Enhances token counter to support the use of session contexts.

```python
tokens_commands = ["**tokens**", "tokens", "token stats", "usage stats"]
```


mhdawson:
The token counts for the Request manager with responses flow look reasonable.

For the request manager flow with non responses I see:

```
📱 App Tokens (from chat agent):
  Input tokens: 62
  Output tokens: 0
  Total tokens: 62
  API calls: 7
```

Which I don't think is right. It's ok if we only get the right counts when using the responses flow, but I'm just asking as I'd have thought these would be 0 if  they were not going to work in the request manager with non-responses flow.

Either way I don't think we should spend too much time on it as I do think we'll be dropping the code for that combination soon and I'm also ok if it only works correctly for the responses flow even if we keep it around.

mhdawson (approved):
Overall LGTM

PR #328, opened 2026-03-02
feat: Session serialization, outbox publish, and dev/test cleanup

tchughesiv:
1. **Session request serialization** – one in-flight request per session, FIFO ordering
2. **Broker publish groundwork** – outbox, advisory locks, durable accept for integration-dispatcher
3. **Dev deps & tests** – standardize dev deps and test targets

---

## 1. Session request serialization

Ensures requests per session are processed in order across request-manager, broker, and agent.

### Request-manager

- Per-session advisory lock; RequestLog (pending → processing → completed/failed)
- Two-phase: durable accept, then reclaim/dequeue/process loop
- Pod heartbeats + background reclaim for stuck requests
- Response poller for cross-pod delivery; polling path forwards to integration-dispatcher when needed
- 503 on lock timeout, RequestLog failure, agent timeout

### Agent

- Per-session lock + RequestLog check (503 if earlier request still pending)
- `agent_received_at` for FIFO verification

### Eventing

- `partitionkey=session_id`; event ID namespacing (`agent-response:{request_id}`); broker-safe IDs for email
- Integration-dispatcher (email): INTERNALDATE sort, claim before forward, Request ID removed from body (Session ID only)

### Migration

- 001: consolidated schema (replaces former 001–004). Pre-production use; no upgrade-path concerns. Old migrations had constraint-name brittleness across envs—004 dropped `processed_events_event_id_key` but DBs had `uq_processed_events_event_id`, so the legacy `UNIQUE(event_id)` survived and broke event claiming. Squash ensures `UNIQUE(event_id, processed_by)` from creation. VARCHAR(255) for `request_id` (email Message-IDs).

---

## 2. Broker publish groundwork

- Shared-models: `outbox.py`, `advisory_lock.py`
- Integration-dispatcher: outbox publish, thread lock for FIFO

---

## 3. Dev deps & tests

- `[dependency-groups] dev` only; `run_pytest` / `run_session_test` macros
- Session integration tests; smoke tests; reclaim helpers

---

## Docs

- `SESSION_SERIALIZATION_RUNBOOK.md` – design, env vars, troubleshooting


agent-service/src/agent_service/main.py
```diff
@@ -42,6 +49,75 @@
...
+    except Exception as e:
+        if event_id:
+            # For retriable errors (503, 502, 504), release the claim so broker retry
+            # can re-claim and process. Marking as "error" would cause try_claim to
+            # skip forever—requests would never complete.
+            is_retriable = False
+            if isinstance(e, HTTPException):
+                is_retriable = e.status_code in (502, 503, 504)
+            if is_retriable:
+                await DatabaseUtils.release_event_claim(db, event_id, "agent-service")
```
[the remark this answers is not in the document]
tchughesiv: That’s a fair question. This retry limit is enforced at the broker level—retry: 10 in `knative-broker.yaml` (and mock eventing uses 10 for 5xx as well). When we return 503, the broker retries; we call `release_event_claim` so each retry can be re-claimed and processed instead of being skipped as a duplicate. Agent-service doesn’t track attempts—the broker decides how many retries we see. If we ever want to fail fast on our side, we can add attempt tracking as a follow-up.

docs/SESSION_SERIALIZATION_RUNBOOK.md
```diff
@@ -0,0 +1,288 @@
...
+- **FIFO** per session by request timestamp.
+- **Sequential agent processing**: The agent processes one request at a time per session (see [How FIFO works end-to-end](#how-fifo-works-end-to-end)).
+- **Durable**: accepted requests stored before processing; no loss on crash (two-phase insert).
+- **Reclaim**: stuck `processing` requests (pod dead or timeout) are reset to `pending` (default) or marked `failed` (configurable).
+
+### Eventing requirements for FIFO
+
+- **Partition key**: `REQUEST_CREATED` and `AGENT_RESPONSE_READY` events use `partitionkey=session_id` so the broker delivers in order per session. Required for sequential agent processing.
+- **`agent_received_at`**: Timestamp when the agent started processing each request. Uses DB time (`SELECT now()`) to avoid pod clock skew across replicas—same approach as `created_at`. Included in responses for FIFO verification; sorting by `agent_received_at` should match send order per session.
+- **Mock eventing**: Must run with 1 replica (`requestManagement.knative.mockEventing.replicas: 1`). Multiple replicas would break per-session FIFO—events for the same `partition_key` could hit different pods.
```
mhdawson, line 17: This means we can only use REPLICA_COUNT=2 in PROD not test, unless we also override the replicas for the mock eventing ?

docs/SESSION_SERIALIZATION_RUNBOOK.md
```diff
@@ -0,0 +1,288 @@
...
+
+### How FIFO works end-to-end
+
+FIFO and sequential agent processing are achieved by two layers working together:
+
+| Layer | Role | Dev branch | This branch |
+|-------|------|------------|-------------|
+| **Request-manager** | FIFO dequeue, sends to broker in order (oldest first) | No: concurrent send, arbitrary order | Yes: per-session lock, RequestLog status, dequeue by `created_at` |
+| **Eventing (partition key)** | Delivers to agent in order per session | No partition key | `partitionkey=session_id` on REQUEST_CREATED, AGENT_RESPONSE_READY |
+| **Agent** | Processes one-at-a-time per partition | Arbitrary order | Sequential: broker queues per partition + per-session DB advisory lock (cross-pod, same approach as request-manager) |
```
mhdawson, line 28: Processes one-at-a-time per partition -> what does a partition map do in terms of the quickstart?

docs/SESSION_SERIALIZATION_RUNBOOK.md
tchughesiv: we override for mock eventing... mock eventing is always 1 now. everything else scales

docs/SESSION_SERIALIZATION_RUNBOOK.md
tchughesiv: we add the following to our broker -
```yaml
  default.topic.partitions: "{{ .Values.requestManagement.knative.broker.config.numPartitions | default 6 }}"
```
and this to our triggers -
```yaml
  annotations:
    kafka.eventing.knative.dev/delivery.order: ordered  # Per-partition FIFO for session ordering
```
it just serializes requests as they flow through and groups them by session. so its a per-session one at a time flow. in the code we now set a "partition key" when we send events that's unique to the session. this provides per-session ordering within each partition.

in mock eventing, we code similar behaviour

guides/PERFORMANCE_SCALING_GUIDE.md
```diff
@@ -42,7 +42,7 @@ Asyncio is a great fit for this quickstart as we've seen that most of the time a
...
+Use standard kubernetes scaling techniques. For agent-service, integration-dispatcher, request-manager, you can scale the number of uvicorn workers as well as the number of replicas. **Mock-eventing must stay at 1 replica** when using partition-key ordering—multiple replicas would break per-session FIFO (events for the same partition_key can hit different pods). For MCP servers (such as snow) the same is true, but special considerations which are covered in a later section. Both the number of workers and number of replicas can be configured in helm/values.yaml. As an example this snippet which is part of the configuration for the agent service sets 4 workers and 2 replicas:
```
mhdawson:
```suggestion
Use standard kubernetes scaling techniques. For agent-service, integration-dispatcher, request-manager, you can scale the number of uvicorn workers as well as the number of replicas. For MCP servers (such as snow) the same is true, but special considerations which are covered in a later section. Both the number of workers and number of replicas can be configured in helm/values.yaml. As an example this snippet which is part of the configuration for the agent service sets 4 workers and 2 replicas:
```

guides/PERFORMANCE_SCALING_GUIDE.md
mhdawson:
```suggestion

In test mode **Mock-eventing must stay at 1 replica** when using partition-key ordering—multiple replicas would break per-session FIFO (events for the same partition_key can hit different pods). 
```

docs/SESSION_SERIALIZATION_RUNBOOK.md
mhdawson: Thanks, see that in the Makefile now.

shared-models/src/shared_models/request_log.py
```diff
@@ -0,0 +1,56 @@
+"""RequestLog ordering utilities for session FIFO.
+
+Used by agent-service to enforce RequestLog as source of truth for processing order
+(defense in depth alongside partition key). Before processing, the agent checks
```
mhdawson:
```suggestion
Before processing, the agent checks
```

mhdawson (approved):
LGTM with a minor suggestion.

PR #343, opened 2026-03-20
feat: support for multiple flows for evaluations

mhdawson:
- add support to have multiple flows each with their own predefined, known bad and generated conversations as well as their own evaluation metrics
- add initial set of known_bad_conversations for ticket-laptop-refresh flow
- existing evaluations are considered the "default" and are unchanged. Running without new flow option will run evaluations in exacltly the same way as before

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -371,6 +371,7 @@ help:
 	@echo "  test-mock-employee-data            - Run tests for mock employee data"
 	@echo "  test-mock-servicenow               - Run tests for mock ServiceNow"
 	@echo "  test-short-resp-integration-request-mgr - Run short responses integration tests with Request Manager"
+	@echo "  test-short-ticket-laptop-refresh   - Run short responses integration tests for ticket-laptop-refresh flow"
 	@echo "  test-long-resp-integration-request-mgr - Run long responses integration tests with Request Manager"
 	@echo "  test-long-concurrent-integration-request-mgr - Run long concurrent responses integration tests with Request Manager (concurrency=4)"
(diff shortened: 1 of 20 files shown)


evaluations/flows/ticket_laptop_refresh/context/.gitkeep
tchughesiv, line 1: what's the reason for this?

mhdawson: I had wondered about that as well and search says:

"The .gitkeep file is an unofficial convention used in Git repositories to force Git to track an otherwise empty directory. 
"

So I think its the typical way to ensure that an empty directory exists after checkout even if there is nothing in it.

mhdawson: If there is a better way, happy to do that as well.

tchughesiv: understood, but i guess my question is why do we need an empty dir here?

mhdawson: The eval code copies files into the directory when it runs and my current assumption is that it needs the empty directory to exist. I'll double check that. Either way I can update the code so that it can create it instead as I don't see us keeping an empty directory in git in other similar cases.

mhdawson:
@tchughesiv removed the .gitkeep, seems the existing code could already cope with the directory not being there.

tchughesiv (approved):
lgtm

PR #345, opened 2026-03-24
fix: add LLM_ID and LLM to for evals in workflows

mhdawson:


mhdawson:
Not sure if we'll want to leave this with the shorter run, but for now with the current token limits I think it makes sense to do a shorter run.

tchughesiv (approved):
lgtm

PR #346, opened 2026-03-26
feat: Zammad MCP server

Prudhvivuda:
**MCP server which securely accepts authoritative user id in the format {user email}-{ticket nubmer}.** 

It provides the following functions
- mark_as_agent_managed_laptop_refresh
- close
- route_to_human_managed_queue
- escalate_for_human_review
- send_to_manager_review


mhdawson:
@Prudhvivuda seems to fail to deploy causing the e2e tests to fail can you please investigate?

Also wondering what kind of testing you did so far?

Prudhvivuda:
> @Prudhvivuda seems to fail to deploy causing the e2e tests to fail can you please investigate?
> 
> Also wondering what kind of testing you did so far?

Sure, I will investigate the reasons for failure. I ran local make tests. I will reach out to you for more details.

tchughesiv (commented):
@Prudhvivuda thanks for the PR... is your approach here leveraging/wrapping the existing zammad mcp server?
https://github.com/rh-ai-quickstart/it-self-service-agent/blob/48c57c0907bd25f13fb8d8de82d8b77544a43071/helm/values.yaml#L272-L286

i assume you're using `helm-install-ticketing` to deploy the stack with zammad and the mcp server?

we don't want to duplicate existing functionality. would be good to thoroughly research what https://github.com/basher83/Zammad-MCP already does if you haven't already... maybe we don't even need a wrapper (we may though).

helm/templates/external-routes.yaml
```diff
@@ -53,3 +53,52 @@ spec:
     termination: edge
     insecureEdgeTerminationPolicy: Redirect
 {{- end }}
+
+{{- if and .Values.zammad.enabled .Values.zammad.externalRoute.enabled }}
+---
```
mhdawson: I think this might be an unintended change due to rebase?

helm/zammad-embed/templates/configmap.yaml
```diff
@@ -35,12 +35,16 @@ data:
     </head>
     <body>
       <h1>Zammad Chat - Embed Code</h1>
-      <p>Copy the following code into your website before the closing <code>&lt;/body&gt;</code> tag:</p>
+      <p>Insert the widget code into the source code of every page the chat shall be visible on. Place it before the closing <code>&lt;/body&gt;</code> tag (or copy the snippet below).</p>
```
mhdawson: I think this might be an unintended change due to rebase?

helm/zammad/templates/bootstrap-job.yaml
```diff
@@ -39,7 +39,7 @@ spec:
           command: ["python3", "/app/bootstrap.py"]
           env:
             - name: ZAMMAD_BASE_URL
-              value: "http://zammad-nginx:8080"
+              value: "http://zammad-stack-nginx:8080"
```
mhdawson: I think all changes in this PR under zammad might be due to rebase instead of being intended as part of quickstart ? -

zammad-bootstrap/bootstrap.py
```diff
@@ -5,9 +5,23 @@
 so this script has no dependency on an externally provisioned token.
 
 Required env vars:
-  ZAMMAD_BASE_URL       e.g. http://zammad-nginx:8080
+  ZAMMAD_BASE_URL       e.g. http://zammad-stack-nginx:8080 (in-cluster nginx service for release zammad-stack)
```
mhdawson: again don't think this should show as a change in this PR

mhdawson:
Overall I think it needs some cleanup to just show the changes intended for this PR.

mcp-servers/zammad/src/zammad_mcp/zammad_auth_id.py
```diff
@@ -0,0 +1,48 @@
+"""Parse composite authoritative user id: {email}-{ticket_id} for Zammad."""
```
mhdawson: I think this file should be in src/zammad_mcp as I don't see why it would be in the mcp_common directory, particularly given its name.

Prudhvivuda: Agreed - moved zammad_auth_id into zammad_mcp and dropped it from mcp-common

helm/Chart.yaml
```diff
@@ -19,7 +19,8 @@ dependencies:
   - name: llama-stack
     version: 0.7.0
     repository: https://rh-ai-quickstart.github.io/ai-architecture-charts
+  # Vendored from ai-architecture-charts 0.5.8 + basherSidecar second container (upstream chart ignores basherSidecar).
   - name: mcp-servers
     version: 0.5.8
-    repository: https://rh-ai-quickstart.github.io/ai-architecture-charts
+    repository: file://charts/mcp-servers
```
tchughesiv: this is a problem... rather than merge like this, we should submit a PR to `ai-architecture-charts` adding support for sidecar/multi-container mcp pods.

tchughesiv (changes requested):

mcp-servers/snow/src/snow/servicenow/headers.py
```diff
@@ -1,71 +1,25 @@
-"""Header extraction utilities for ServiceNow MCP server."""
+"""ServiceNow-specific MCP headers (SERVICE_NOW_TOKEN). Shared helpers in mcp_common."""
```
mhdawson: I don't think this file should be needed at all ? server.py could just call api_token = header_first(ctx, "SERVICE_NOW_TOKEN")  directly instead ?

Prudhvivuda: Removed servicenow/headers.py; Snow now calls mcp_common.headers.header_first for SERVICE_NOW_TOKEN and keeps the ticket-suffix strip as a small private helper in server.py since that behavior is Snow-only.

helm/Chart.yaml
Prudhvivuda: Makes sense. Submitted a PR to `ai-architecture-charts`

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,157 @@
+"""Call the in-pod Basher Zammad MCP (streamable HTTP) from sync tool handlers.
+
+Runs each request in a worker thread with its own asyncio loop so we never call
+``asyncio.run`` from FastMCP's event loop thread.
+
+Basher (basher83/Zammad-MCP) authenticates using **only** process environment: ``ZAMMAD_URL``
+(with ``/api/v1`` as in upstream ``.env.example``), ``ZAMMAD_HTTP_TOKEN`` or
+``ZAMMAD_HTTP_TOKEN_FILE``, or OAuth2 / password per upstream — not MCP headers from this
+wrapper. Legacy ``ZAMMAD_API_TOKEN`` env is also common in our Helm charts.
```
mhdawson: Not sure what this comment mean - not MCP headers from this
wrapper. Legacy ``ZAMMAD_API_TOKEN`` env is also common in our Helm charts.

Overall not sure what this comment is trying to help us understand.

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,320 @@
...
+import os
+from typing import Any, Dict, Optional
+
+import httpx
+from shared_models import configure_logging
+
+logger = configure_logging("zammad-mcp-server")
+
+
+def _read_token_from_file(env_var: str) -> str:
```
mhdawson: Why is this needed ? I think its passed in through an Environment Variable right ?

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,320 @@
...
+            path=path,
+            error=str(e),
+        )
+        return ""
+
+
+def _env_zammad_token() -> str:
+    """Token from wrapper process env / *_FILE mounts (Basher-compatible order)."""
+    return (
+        _read_token_from_file("ZAMMAD_HTTP_TOKEN_FILE")
```
mhdawson: Why is this trying different things? We know the environment variable so we should be able to just read the right one right?

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,320 @@
...
+    return base
+
+
+class ZammadClient:
+    """Minimal Zammad REST client: ticket + user fetch for AUTHORITATIVE_USER_ID checks."""
+
+    def __init__(self, base_url: str, api_token: str) -> None:
+        self._base = normalize_zammad_base_url(base_url)
+        self._timeout = _rest_timeout_seconds()
+        self._verify = _rest_tls_verify()
```
mhdawson: do we really need to do this?

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,320 @@
...
+    if not api_token:
+        raise ValueError(
+            "Zammad API token is required: set ZAMMAD_HTTP_TOKEN (Basher canonical), "
+            "ZAMMAD_HTTP_TOKEN_FILE, ZAMMAD_API_TOKEN env, or MCP headers — see resolve_api_token."
+        )
+    base = normalize_zammad_base_url(raw)
+    return ZammadClient(base_url=base, api_token=api_token)
+
+
+def resolve_api_token(ctx_token: Optional[str]) -> str:
```
mhdawson: Can you explain in a bit more detail why this is needed ? Since the container is inside the same pod I'm wondering if its really needed ?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,363 @@
...
+    try:
+        client, _email, ticket_id = _authorize_and_client(ctx)
+        field = _env_str("ZAMMAD_USER_MANAGER_FIELD", "manager_email")
+        user = _ticket_customer_from_basher(ticket_id)
+        raw = user.get(field)
+        manager = (str(raw).strip() if raw is not None else "") or ""
+        if not manager:
+            manager = (
+                os.getenv("ZAMMAD_MANAGER_EMAIL")
+                or os.getenv("ZAMMAD_MANAGER_LOGIN")
```
mhdawson: 1) not sure why we'd be checking two different things here. This is typical AI generated stuff where it tries a bunch of different things until it finds the righ one.

Could do go through all of the new code and remove cases like this. We know what it should be so we should just check the one. 

2) I also think that putting all of the code that gets values from the environment into one function. Then it wouldjust be one line like 

if not manager:
  manager = config.default_manager

I think doing that everywhere would make the code a lot more readable.

Makefile
```diff
@@ -1522,6 +1581,8 @@ define helm_install_common
 		--set-string mcp-servers.mcp-servers.self-service-agent-snow.env.SERVICENOW_LAPTOP_REQUEST_LIMITS="$(SERVICENOW_LAPTOP_REQUEST_LIMITS)" \
 		--set mcp-servers.mcp-servers.self-service-agent-snow.envSecrets.SERVICENOW_INSTANCE_URL.name=$(MAIN_CHART_NAME)-servicenow-credentials \
 		--set mcp-servers.mcp-servers.self-service-agent-snow.envSecrets.SERVICENOW_INSTANCE_URL.key=servicenow-instance-url \
+		--set mcp-servers.mcp-servers.zammad-mcp.image.repository=$(REGISTRY)/self-service-agent-zammad-mcp \
```
mhdawson: I think these should only be set if ZAMMAD is enabled ?

Makefile
```diff
@@ -1580,6 +1641,8 @@ define helm_export
...
```
mhdawson: I think these should only be set if ZAMMAD is enabled ?

Makefile
```diff
@@ -906,7 +935,7 @@ reinstall:
 	@echo "All dependencies reinstalled with latest code!"
 
 .PHONY: reinstall-all
-reinstall-all: reinstall-shared-models reinstall-shared-clients reinstall reinstall-request-manager reinstall-agent-service reinstall-integration-dispatcher reinstall-mcp-snow reinstall-mock-employee-data reinstall-mock-servicenow reinstall-promptguard reinstall-servicenow-bootstrap
+reinstall-all: reinstall-shared-models reinstall-shared-clients reinstall reinstall-request-manager reinstall-agent-service reinstall-integration-dispatcher reinstall-mcp-snow reinstall-mcp-zammad reinstall-mock-employee-data reinstall-mock-servicenow reinstall-promptguard reinstall-servicenow-bootstrap
```
mhdawson, line 941: I wonder if we can make sure this only has an effect if ZAMMAD is enabled?

mhdawson: same in deps-all above

Makefile
```diff
@@ -1,6 +1,6 @@
...
+ifneq (,$(filter namespace install uninstall helm-install-test helm-install-prod helm-install-demo helm-install-ticketing ticketing-clear-ss-conflicts helm-export-demo helm-export-validate-demo ansible-apply-demo ansible-teardown-demo helm-uninstall helm-status helm-cleanup-eventing helm-cleanup-jobs deploy-email-server undeploy-email-server deploy-zammad undeploy-zammad zammad-set-token zammad-bootstrap-token zammad-trigger-autowizard print-urls verify-triggers jaeger-deploy jaeger-undeploy test-session-serialization-integration test-session-reclaim-integration test-session-background-reclaim-integration,$(MAKECMDGOALS)))
```
mhdawson: wondering why the change to add ticketing-clear-ss-conflicts ? Is that part of the MCP changes ?

mcp-servers/zammad/src/zammad_mcp/basher_client.py
Prudhvivuda: Simplified the module docstring to describe what this file does only.

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,365 @@
...
+            raise
+        msg = f"close failed: {_tool_error_text(e)}"
+        logger.exception(msg)
+        return msg
+
+
+@mcp.tool()
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
+def escalate_for_human_review(ctx: Context[Any, Any], note: str = "") -> str:
+    """Escalate this ticket to the human escalation queue (tag + group only).
```
mhdawson: I'd shrunk the descriptions to remove the amount of tokens going into the agent context.
Can you pull those changes in as well. For example in this case I had reduced it to just

    """Escalate this ticket to the human escalation queue."""

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: removed others and left base URL normalisation because Helm stores zammad-url with a trailing /api/v1.

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: Cursor says - We don’t know at build time which key a given Secret or local .env uses. Basher uses ZAMMAD_HTTP_TOKEN / ZAMMAD_HTTP_TOKEN_FILE; we keep ZAMMAD_API_TOKEN as a pod fallback so it matches the legacy header name and older setups. Order is fixed (file, then HTTP, then API) - first non-empty wins.

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: Basher only reads pod env, not MCP HTTP headers. The MCP client runs outside the pod and may send ZAMMAD_HTTP_TOKEN / ZAMMAD_API_TOKEN on the request (local dev, tests). resolve_api_token decides what the wrapper uses for its own Zammad REST auth gate: pod-only to match Basher (ZAMMAD_WRAPPER_USE_POD_TOKEN_ONLY), header when there is no pod token, or optional header↔pod match checks. Same-pod Secret does not remove that choice because headers are a separate channel Basher never sees.

Makefile
Prudhvivuda: ticketing-clear-ss-conflicts runs kubectl against the namespace (cleans up objects that block Helm SSA on re-install). It was added to the same NAMESPACE guard as the other cluster targets so you can’t run it with NAMESPACE unset. It’s not part of MCP changes

Makefile
```diff
@@ -1659,6 +1722,14 @@ helm-install-ticketing: namespace helm-depend
 	@echo "Step 4/4: Printing checklist..."
 	@$(MAKE) _helm-install-ticketing-print-checklist NAMESPACE=$(NAMESPACE)
 
+# One-shot fix when helm upgrade fails with server-side apply conflicts on ticketing resources.
+.PHONY: ticketing-clear-ss-conflicts
+ticketing-clear-ss-conflicts: namespace
+	@echo "Removing objects that commonly conflict with Helm SSA (they are recreated on the next upgrade)."
+	kubectl delete configmap $(MAIN_CHART_NAME)-zammad-embed -n $(NAMESPACE) --ignore-not-found
+	kubectl delete deployment mcp-zammad-mcp -n $(NAMESPACE) --ignore-not-found
+	@echo "Done. Re-run: make helm-install-ticketing NAMESPACE=$(NAMESPACE)"
```
tchughesiv: this shouldn't be necessary... embed-related objects are fully cleaned up now and the following PR will further improve.
https://github.com/rh-ai-quickstart/it-self-service-agent/pull/357

wrt the zammad mcp deployment, that should be cleaned up automatically as part of the helm-uninstall right? no need for an extra command

mcp-servers/zammad/src/zammad_mcp/server.py
Prudhvivuda: okay

Makefile
Prudhvivuda: agreed and dropped ticketing-clear-ss-conflicts

helm/values.yaml
```diff
@@ -268,22 +268,61 @@ mcp-servers:
     oracle-sqlcl:
       enabled: false
 
-    # Zammad MCP server for ticketing integration
+    # Zammad MCP wrapper (AUTHORITATIVE_USER_ID gate + in-pod Basher Zammad-MCP for mutations)
```
mhdawson:
```suggestion
    # Zammad MCP server for ticketing integration
```

helm/values.yaml
```diff
@@ -268,22 +268,61 @@ mcp-servers:
...
+        repository: quay.io/rh-ai-quickstart/self-service-agent-zammad-mcp
+        tag: "0.0.12"
       port: 8000
       targetPort: 8000
-      transport: http
-      imagePullPolicy: IfNotPresent
+      transport: streamable-http
+      imagePullPolicy: Always
+      # Basher Zammad MCP (AGPL) in the same pod; wrapper calls 127.0.0.1. Requires mcp-servers chart >=0.5.17 (extraContainers).
+      # Secret name must match ZAMMAD_CREDENTIALS_SECRET / Makefile MAIN_CHART_NAME (default install: self-service-agent).
```
mhdawson:
```suggestion
```

helm/values.yaml
```diff
@@ -268,22 +268,61 @@ mcp-servers:
...
+              value: "8001"
       env:
-        MCP_TRANSPORT: "http"
-        MCP_PORT: "8000"
-        # ZAMMAD_URL and ZAMMAD_HTTP_TOKEN from Secret (zammad.url, zammad.mcp.token)
+        MCP_TRANSPORT: "streamable-http"
+        UVICORN_WORKERS: "2"
+        # No trailing slash: avoids some Basher/streamable-http 307 redirect quirks (see basher_client).
+        ZAMMAD_BASHER_MCP_URL: "http://127.0.0.1:8001/mcp"
+        # Wrapper REST auth uses the same Secret token as Basher (ignore Cursor header overrides).
```
mhdawson:
```suggestion
        # Wrapper REST auth uses the same Secret token as Basher
```

helm/values.yaml
```diff
@@ -268,22 +268,61 @@ mcp-servers:
...
+        ZAMMAD_WRAPPER_USE_POD_TOKEN_ONLY: "1"
+        # ZAMMAD_URL and ZAMMAD_HTTP_TOKEN from Secret (zammad-url, zammad-http-token);
+        # Optional: ZAMMAD_REQUIRE_HEADER_MATCHES_POD=1 to reject header≠pod when not using pod-only.
+        # Groups/users from zammad-bootstrap (Michael’s seeding); override if your Zammad differs.
```
mhdawson:
```suggestion
```

mcp-servers/mcp-common/src/mcp_common/headers.py
```diff
@@ -0,0 +1,56 @@
...
+
+def extract_authoritative_user_id(ctx: Context[Any, Any]) -> str | None:
+    """Extract AUTHORITATIVE_USER_ID from MCP request headers.
+
+    Used by Zammad MCP (full ``{email}-{ticket_id}``). The Snow MCP server strips
+    a trailing ``-{ticket_id}`` suffix locally before ServiceNow email lookup.
+
+    If there is no HTTP request (e.g. **stdio** transport), headers are empty. In
+    that case falls back to the process environment variable ``AUTHORITATIVE_USER_ID``
+    (set it in your MCP client config, e.g. Cursor ``env``).
```
mhdawson: Why do we we need this?

mcp-servers/mcp-common/src/mcp_common/headers.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    raw = header_first(ctx, "AUTHORITATIVE_USER_ID", "authoritative_user_id")
+    if raw is not None:
+        return raw
+    env_val = os.getenv("AUTHORITATIVE_USER_ID", "").strip()
```
mhdawson: Same question as above

mcp-servers/snow/src/snow/server.py
```diff
@@ -94,7 +103,7 @@ async def health(request: Any) -> JSONResponse:
 
 
 @mcp.tool()
-@trace_mcp_tool()
+@trace_mcp_tool()  # type: ignore[misc]
```
mhdawson: wondering why this is needed now? the type ignore?

mcp-servers/snow/src/snow/server.py
```diff
@@ -209,7 +218,7 @@ def open_laptop_refresh_ticket(
...
```
mhdawson: same question here

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,141 @@
...
+    arguments: dict[str, Any] = {"params": params}
+
+    def _run() -> str:
+        try:
+            return asyncio.run(_call_tool_async(url, name, arguments))
+        except BaseException as e:
+            formatted = _format_basher_failure(url, name, e)
+            logger.exception("Basher MCP call failed: %s", formatted)
+            if is_basher_unreachable_error(e):
+                raise RuntimeError(f"Basher MCP unreachable: {formatted}") from e
```
mhdawson: wondering why handle this one separately? Do we do something different than other failures?

mcp-servers/mcp-common/src/mcp_common/headers.py
Prudhvivuda: stdio MCP has no HTTP request, so there are no headers. We still need email-ticketId for auth, so we read AUTHORITATIVE_USER_ID from the process env (e.g. Cursor env, local tests). streamable-http keeps using the header path when the request exists.

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,293 @@
...
+                "they override pod env unless ZAMMAD_WRAPPER_USE_POD_TOKEN_ONLY=1. "
+                "Basher always uses the pod token — align them or use pod-only mode."
+            ) from e
+        raise
+
+
+def normalize_zammad_base_url(raw: str) -> str:
+    """Strip trailing ``/api/v1`` segments so paths like ``/api/v1/tickets`` are not doubled.
+
+    Helm/Makefile store ``zammad-url`` as ``http://zammad-nginx...:8080/api/v1``;
```
mhdawson: Is there some reason we can't just set the URL correctly in the makefile and remove this code ?

mcp-servers/mcp-common/src/mcp_common/headers.py
Prudhvivuda: Same reason as above - stdio has no HTTP headers, so header_first is always empty. Env is how Cursor / local dev / tests pass AUTHORITATIVE_USER_ID. In-cluster HTTP still uses the header when present.

mcp-servers/snow/src/snow/server.py
Prudhvivuda: `mypy` on the Snow package is treating `@trace_mcp_tool()` as an untyped decorator, so it reports that the decorated tool becomes untyped ([misc] / “untyped decorator…”). The `# type: ignore[misc]` silences that without changing runtime behavior.

mcp-servers/snow/src/snow/server.py
Prudhvivuda: `mypy` on the Snow package is treating `@trace_mcp_tool()` as an untyped decorator, so it reports that the decorated tool becomes untyped ([misc] / “untyped decorator…”). The `# type: ignore[misc]` silences that without changing runtime behavior.

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,358 @@
...
+        f"Basher {what} did not return a JSON object (expected response_format=json)."
+    )
+
+
+def _basher_add_tag(ticket_id: int, tag: str) -> None:
+    call_basher_tool("zammad_add_ticket_tag", {"ticket_id": ticket_id, "tag": tag})
+
+
+def _update_ticket_fields(ticket_id: int, fields: dict[str, Any]) -> None:
+    call_basher_tool("zammad_update_ticket", {"ticket_id": ticket_id, **fields})
```
mhdawson: why these one liners instead of calling in-line ?

mcp-servers/zammad/src/zammad_mcp/basher_client.py
Prudhvivuda: Same handling. We raise RuntimeError with the same details. The if is_basher_unreachable_error bit only changes the wording (unreachable vs failed) so logs show network/sidecar problems vs errors after connecting. Happy to collapse to a single RuntimeError prefix if we'd rather keep one string.

mcp-servers/zammad/src/zammad_mcp/settings.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _laptop_specialist_owner() -> str | None:
+    if "ZAMMAD_LAPTOP_SPECIALIST_OWNER" not in os.environ:
+        return "agent.laptop-specialist@example.com"
+    v = os.environ["ZAMMAD_LAPTOP_SPECIALIST_OWNER"].strip()
+    return v if v else None
+
+
+def _read_token_file_contents(path: str) -> str:
```
mhdawson: wondering what the Zammad token file is?

mcp-servers/zammad/src/zammad_mcp/settings.py
```diff
@@ -0,0 +1,174 @@
...
+    except OSError as e:
+        _logger.warning(
+            "Could not read Zammad token file",
+            path=p,
+            error=str(e),
+        )
+        return ""
+
+
+def _zammad_url_raw() -> str | None:
```
mhdawson: what is the difference between the URL and the BASE_URL ?

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: We could tighten Makefile/Helm, but ZAMMAD_URL isn’t only from the Makefile, same Secret/env is shared with Basher (upstream often uses `…/api/v1`), plus local .env, other charts, and copy-paste. `normalize_zammad_base_url` avoids `/api/v1/api/v1` on REST paths no matter which source got a trailing `/api/v1`.

mhdawson:
To reduce the direct API calls to the bare minimum suggestion from Claude ->

● Yes, correct. With restructuring:                         
                                                                                                                                                            
  - get_ticket in verify_ticket_customer_email: could call zammad_get_ticket via Basher to get customer_id, then only call direct REST get_user(customer_id)
   to fetch the full user with custom fields and verify the email.
  - get_user_id_by_email: could call zammad_search_users via Basher with the email string, parse the JSON result, and match on email to extract id — since  
  both fields survive Basher's User model.                                                                                                                  
                                                                                                                                                            
  That would leave get_user as the only direct REST call, purely because it's the one place that genuinely needs the custom fields that Basher drops. The   
  ZammadClient could shrink to just get_user plus the shared URL normalization and auth logic.

mhdawson:
Unused code that we can likely remove:
                                                                                                                                                          
  1. _loads_json_dict in server.py (lines 64–121) — 57 lines of JSON-parsing logic that was previously used by the old _ticket_customer_from_basher function
   (removed in the ce08eaf rebase). It has no callers in production code now. The tests in test_tool_errors.py only exist to test it, so they'd go too.     
                                                                                                                                                            
  2. MCP_PORT in server.py (line 180) — assigned from load_mcp_listen_env().port but never passed to FastMCP() or mcp.run(). It's a no-op assignment.

mhdawson:
And another suggestion to reduce the total lines of code, and improves what gets recorded when tracing is enabled and Exceptions occur

Extract @_handle_tool_errors decorator (~26 lines net saved)                                                                                              
                                                            
  In server.py every tool (mark_as_agent_managed_laptop_refresh, close, escalate_for_human_review, send_to_manager_review, route_to_human_managed_queue,    
  get_employee_laptop_info) has an identical 7-line boilerplate block:                                                                                      
      try:                                                                                                                                                  
          ...                                               
      except BaseException as e:                                                                                                                            
          if isinstance(e, (KeyboardInterrupt, SystemExit)):                                                                                                
              raise                                                                                                                                         
          msg = f"<tool_name> failed: {_tool_error_text(e)}"                                                                                                
          logger.exception(msg)                                                                                                                             
          return msg                                                                                                                                        
                                                                                                                                                            
  Replace this with a single decorator in server.py:                                                                                                        
  @functools.wraps(func)                                                                                                                                    
  def _handle_tool_errors(func):                                                                                                                            
      def wrapper(*args, **kwargs):                                                                                                                         
          try:                                                                                                                                              
              return func(*args, **kwargs)                                                                                                                  
          except Exception as e:                                                                                                                            
              msg = f"{func.__name__} failed: {_tool_error_text(e)}"                                                                                        
              logger.exception(msg)                                                                                                                         
              return msg                                                                                                                                    
      return wrapper                                                                                                                                        
                                                                                                                                                            
  except Exception replaces except BaseException + the isinstance(e, (KeyboardInterrupt, SystemExit)) re-raise — KeyboardInterrupt and SystemExit are not   
  subclasses of Exception so they propagate naturally.                                                                                                      
                                                            
  Apply it to each tool above @trace_mcp_tool() so that exceptions are recorded as StatusCode.ERROR in traces before being caught:                          
  @mcp.tool()                                               
  @_handle_tool_errors                                                                                                                                      
  @trace_mcp_tool()                                                                                                                                         
  def close(...):

mcp-servers/mcp-common/src/mcp_common/headers.py
```diff
@@ -0,0 +1,56 @@
...
+    if headers is None:
+        return None
+    for name in names:
+        value = headers.get(name)
+        if value is not None:
+            return cast(str, str(value))
+    return None
+
+
+def extract_authoritative_user_id(ctx: Context[Any, Any]) -> str | None:
```
mhdawson: Should this be here if only used by the zammad MCP server? 

It might be replaced by a one line call to 

raw = header_first(ctx, "AUTHORITATIVE_USER_ID", "authoritative_user_id")

In _authorize_and_client in mcp-servers/zammad/src/zammad_mcp/server.py

mcp-servers/mcp-common/README.md
```diff
@@ -0,0 +1,8 @@
+# mcp-common
+
+Shared Python utilities for MCP servers in this repository:
+
+- **`mcp_common.headers`** — `AUTHORITATIVE_USER_ID` extraction (Snow, Zammad, etc.)
+- **`mcp_common.tracing`** — OpenTelemetry `trace_mcp_tool` decorator
+
+Add as a path dependency from `mcp-servers/snow` or `mcp-servers/zammad`.
```
mhdawson:
```suggestion
```

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,141 @@
+"""Call the Basher Zammad MCP sidecar over streamable HTTP (runs MCP work on a worker thread so FastMCP's loop stays safe)."""
```
mhdawson:
```suggestion
"""Call the Basher Zammad MCP sidecar over streamable HTTP.                                                                                               
                                                            
The Basher MCP client is async-only (MCP SDK has no sync equivalent). Tool handlers                                                                       
are sync def functions running in FastMCP's anyio thread pool - no event loop is                                                                          
present on those threads. The executor provides a guaranteed clean thread where                                                                           
asyncio.run() can start a fresh event loop to drive the async Basher client.                                                                              
                                                                                                                                                            
Tool handlers cannot currently be async def because trace_mcp_tool's inner wrapper
is a sync def - awaiting an async tool would silently return a coroutine object instead of                                                                          
executing it.                                                                                                                                             
"""      ```

File: mcp-servers/zammad/src/zammad_mcp/basher_client.py [hunk 36]
```diff
@@ -0,0 +1,141 @@
...
+from mcp.types import CallToolResult, TextContent
+from shared_models import configure_logging
+from zammad_mcp.settings import load_basher_connect_env
+
+logger = configure_logging("zammad-mcp-basher")
+
+_executor = ThreadPoolExecutor(max_workers=8, thread_name_prefix="zammad-basher-mcp")
+
+
+def basher_mcp_url() -> str:
```
mhdawson: Can this not just be in settings.py so that the value can just be re-used from the settings loaded at the start?

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,141 @@
...
+    if isinstance(sc, list):
+        return json.dumps(sc, default=str)
+    if sc is not None:
+        tail = str(sc)
+        return f"{text}\n{tail}".strip() if text else tail
+    return text or "(no content)"
+
+
+async def _call_tool_async(url: str, name: str, arguments: dict[str, Any]) -> str:
+    timeout = httpx.Timeout(120.0, read=120.0)
```
mhdawson: We should make the timeout configurable through a variable set as part of the deployment

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,141 @@
...
+                parts.append(block.text)
+        msg = " ".join(parts).strip() or "unknown error"
+        raise RuntimeError(msg)
+    text_blocks: list[str] = []
+    for block in result.content:
+        if isinstance(block, TextContent):
+            text_blocks.append(block.text)
+    text = "\n".join(t for t in text_blocks if t).strip()
+    sc = result.structuredContent
+    # Dict/list structured payloads: return JSON only. Mixing text + JSON breaks callers that
```
mhdawson: The comment on line 83-84 is now misleading: it says "Mixing text + JSON breaks callers that json.loads the full string (e.g. zammad_get_ticket with response_format=json)" — that's referring to       
  _loads_json_dict, which is dead code. So the justification for the structuredContent handling is gone.                                                                                                  
                                                                                                                                                                                                          
  The structuredContent block (lines 82-92, 8 lines) may now be unnecessary. None of the current tools (zammad_add_ticket_tag, zammad_update_ticket, zammad_add_article) use the return value of          
  call_basher_tool for anything — they don't parse it. So whether structured content is serialised to JSON or not makes no difference.                                                                    
                                                                                                                                                                                                          
  The text extraction is duplicated between the error path (lines 71-75) and success path (lines 77-81) — both iterate result.content filtering TextContent. That's ~4 lines of duplication that could be 
  extracted into a one-liner like:
  " ".join(b.text for b in result.content if isinstance(b, TextContent)).strip()                                                                                                                          
                                                                                
  So combined — removing the structuredContent block and deduplicating the text extraction — saves roughly 10-12 lines and removes the now-misleading comment.

mcp-servers/zammad/src/zammad_mcp/client.py
mhdawson: or if necessary move it to settings.py

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,293 @@
...
+        base = base[: -len(suffix)].rstrip("/")
+    return base
+
+
+class ZammadClient:
+    """Minimal Zammad REST client: ticket + user fetch for AUTHORITATIVE_USER_ID checks.
+
+    ``__init__`` strips repeated ``/api/v1`` from *base_url* so request paths (which always
+    include ``/api/v1/...``) are not doubled when ``ZAMMAD_URL`` comes from Helm with a trailing
+    ``/api/v1``. Uses a fixed REST timeout and httpx default TLS verification.
```
mhdawson:
```suggestion
```

mcp-servers/zammad/src/zammad_mcp/headers.py
```diff
@@ -0,0 +1,29 @@
...
+from mcp_common.headers import extract_authoritative_user_id, header_first
+
+__all__ = ["extract_authoritative_user_id", "extract_zammad_api_token"]
+
+
+def extract_zammad_api_token(ctx: Context[Any, Any]) -> str | None:
+    """Extract Zammad API token from MCP HTTP headers (optional pass-through).
+
+    Names align with **basher83/Zammad-MCP** (``ZAMMAD_HTTP_TOKEN`` in ``.env``) and legacy
+    ``ZAMMAD_API_TOKEN``. Starlette header lookup is case-insensitive for the canonical forms.
```
mhdawson: This is brand new code there is no need to support "legacy" names.

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,290 @@
...
+        "(same as Basher)."
+    )
+
+
+def _customer_id_from_ticket(ticket: Dict[str, Any]) -> Optional[int]:
+    """Resolve Zammad ticket customer id from ``customer_id`` or expanded ``customer`` object."""
+    raw = ticket.get("customer_id")
+    if raw is not None:
+        return int(raw)
+    cust = ticket.get("customer")
```
mhdawson: Not sure why we are calling twice with different names, we should just call once with the right name?

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,290 @@
...
+    try:
+        user = client.get_user(customer_id)
+    except httpx.TransportError as e:
+        raise ValueError(
+            "Zammad REST unreachable loading ticket customer (user). "
+            "Align ZAMMAD_URL with Basher and network policy. "
+            f"{type(e).__name__}: {e}"
+        ) from e
+    # Zammad exposes email and/or login
+    user_email = (user.get("email") or user.get("login") or "").strip().lower()
```
mhdawson: this looks like fallback as well, we should just get the right one.

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def _post_optional_user_note(
```
mhdawson: would _post_user_note be a better name

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,358 @@
...
+        logger.exception(msg)
+        return msg
+
+
+@mcp.tool()
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
+def close(ctx: Context[Any, Any], note: str = "") -> str:
+    """Close the ticket."""
+    try:
+        cfg = load_zammad_mcp_settings()
```
mhdawson: If this is on a class can we do it once for the class so that we only have to get it once.

mcp-servers/zammad/src/zammad_mcp/server.py
Prudhvivuda: done

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: done

mcp-servers/zammad/src/zammad_mcp/headers.py
Prudhvivuda: done

mcp-servers/mcp-common/src/mcp_common/headers.py
Prudhvivuda: moved to `mcp-servers/zammad/src/zammad_mcp/server.py`. `env_val = os.getenv("AUTHORITATIVE_USER_ID", "").strip()` would help when using cursor mcp to test.

mcp-servers/snow/src/snow/server.py
mhdawson: What I was hoping to understand is why it was not an issue before ?

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,67 @@
...
+            return asyncio.run(_call_tool_async(url, name, arguments))
+        except BaseException as e:
+            detail = f"tool={name!r} url={url!r}: {type(e).__name__}: {e}"
+            logger.exception("Basher MCP call failed: %s", detail)
+            raise RuntimeError(f"Basher MCP failed: {detail}") from e
+
+    return _executor.submit(_run).result(timeout=BASHER_MCP_TIMEOUT_SECONDS)
+
+
+def note_body(*, subject: str, default_body: str, user_note: str = "") -> str:
```
mhdawson: This does not seem related to the basher_client.py and would likely be better to be inlined into _post_user_note in server.py

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,203 @@
...
+from typing import Any, Dict
+
+import httpx
+from shared_models import configure_logging
+from zammad_mcp.basher_client import call_basher_tool
+from zammad_mcp.settings import zammad_http_token_from_env, zammad_rest_base_url
+
+logger = configure_logging("zammad-mcp-server")
+
+_REST_TIMEOUT_SEC = 30.0
```
mhdawson: Instead of having another timeout value, lets use the same timeout that is used for the MCP server calls. BASHER_MCP_TIMEOUT_SECONDS.  Maybe we should change that to ZAMMAD_MCP_TIMEOUT_SECONDS.

The both are used for requests against the same server so the same timeout makes sense to me?

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,203 @@
...
+            try:
+                data = r.json()
+            except json.JSONDecodeError as e:
+                raise ValueError(
+                    "Zammad REST returned non-JSON (check ZAMMAD_URL / reverse proxy)."
+                ) from e
+            return dict(data)
+
+
+def read_ticket_state(ticket: Dict[str, Any]) -> tuple[int | None, str]:
```
mhdawson: I think this is no longer used and could be removed?

Prudhvivuda: yes, thanks for pointing. removed

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,203 @@
...
+    def __init__(self, base_url: str, http_token: str) -> None:
+        self._base = base_url.strip().rstrip("/")
+        self._timeout = _REST_TIMEOUT_SEC
+        self._headers = {
+            "Authorization": f"Token token={http_token}",
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+
+    def _url(self, path: str) -> str:
```
mhdawson: I think this could just be inlined into the one placde it is called?

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: done

mcp-servers/zammad/src/zammad_mcp/basher_client.py
Prudhvivuda: yes, moved it.

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: okay

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
...
+            "type": "note",
+            "sender": "Agent",
+            "internal": False,
+        },
+    )
+
+
+@asynccontextmanager
+async def lifespan(app: FastMCP) -> AsyncGenerator[None, None]:
+    base = zammad_mcp_settings.zammad_rest_base_url()
```
mhdawson: this should likely be something like base=zammad_mcp_settings.ZAMMAD_REST_BASE_URL to be consistent, right ?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
...
+    else:
+        logger.info("Zammad MCP wrapper configured", zammad_url=base.rstrip("/"))
+    logger.info("Basher MCP URL", basher_mcp_url=zammad_mcp_settings.BASHER_MCP_URL)
+    try:
+        yield
+    finally:
+        logger.info("Shutting down Zammad MCP server")
+
+
+_mcp_listen = zammad_mcp_settings.load_mcp_listen_env()
```
mhdawson: I settings should mostly be reading from the environment and then the variables used here to create the env.

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
...
+    """Headers + verify; returns (rest_client, email, ticket_id, customer_user from GET /users/{id})."""
+    raw = header_first(ctx, "AUTHORITATIVE_USER_ID", "authoritative_user_id")
+    if not raw:
+        raise ValueError(
+            "AUTHORITATIVE_USER_ID missing on the MCP request (set the tool header from the agent). "
+            "Expected format: {email}-{ticket_id} (Zammad internal ticket id)."
+        )
+    email, ticket_id = parse_email_and_ticket_id(raw)
+    token = get_zammad_api_token()
+    client = client_from_env_and_token(token)
```
mhdawson: I don't understand why we have a method to get the token and then pass it into client_from_env_and_token, could the call to get_zammad_api_token not just be done in the client itself.  It would then likely make sense to have the method called something like get_zammad_rest_client()

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
...
+        parts.append(f"owner set to {owner!r}")
+    return ", ".join(parts) + "."
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
+def close(ctx: Context[Any, Any], note: str = "") -> str:
+    """Close the ticket."""
+    _, _email, ticket_id, _ = _authorize_and_client(ctx)
```
mhdawson: I think this will do a REST lookup when it is not needed ?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
+"""Zammad ticket MCP: AUTHORITATIVE_USER_ID gate; mutations via Basher."""
```
mhdawson:
```suggestion
"""Zammad ticket MCP"""
```

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,271 @@
...
+    finally:
+        logger.info("Shutting down Zammad MCP server")
+
+
+_mcp_listen = zammad_mcp_settings.load_mcp_listen_env()
+MCP_TRANSPORT: Literal["stdio", "sse", "streamable-http"] = _mcp_listen.transport  # type: ignore[assignment]
+mcp = FastMCP(
+    "Zammad Ticket MCP",
+    host=_mcp_listen.host,
+    port=_mcp_listen.port,
```
mhdawson: Could these just be a  zammad_mcp_settings.MCP_LISTEN_HOST, and  zammad_mcp_settings.MCP_LISTEN port or something like that?

mcp-servers/zammad/src/zammad_mcp/settings.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+    zammad_http_token = os.getenv("ZAMMAD_HTTP_TOKEN", "").strip()
+    if not zammad_http_token:
+        raise ValueError(
+            "ZAMMAD_HTTP_TOKEN is required on this process "
+            "(Kubernetes secret zammad-http-token / local export)."
+        )
+
+    basher_raw = os.getenv("ZAMMAD_BASHER_MCP_URL", "").strip().rstrip("/")
+    basher_mcp_url = basher_raw if basher_raw else _DEFAULT_BASHER_MCP_URL
```
mhdawson: To be consistent with the rest of the defaults this should probably just be

    basher_mcp_url = basher_raw if basher_raw else "http://127.0.0.1:8001/mcp"

and then remove _DEFAULT_BASHER_MCP_URL above.

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,185 @@
+"""Basher-first ticket tools."""
```
mhdawson:
```suggestion
"""Client for using the Zammad client API directly. Our goal is to avoid this and
     pass all requests through the MCP server, however that is not possible for custom attributes
     on the user (manager email and current laptop info). as all of the mcp server tools strips those out"""
```

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,185 @@
...
+        if not isinstance(u, dict):
+            continue
+        if str(u.get("email") or "").strip().lower() == want:
+            uid = int(u["id"])
+            logger.info("resolved user id", email=email, user_id=uid)
+            return uid
+    raise ValueError(f"No Zammad user found with email {email!r}.")
+
+
+def assert_ticket_customer_matches_basher(ticket_id: int, expected_email: str) -> int:
```
mhdawson: I think both get_user_id_by_email and assert_ticket_customer_matches_basher should be moved to basher_client.py                     
                                                                                                  
both only use call_basher_tool, neither touches the REST client, so moving both to basher_client.py works cleanly

Gives a clear module boundary: basher_client.py owns all Basher MCP interactions, client.py owns
  all direct Zammad REST interactions.

mcp-servers/zammad/src/zammad_mcp/client.py
mhdawson: I think it would also be good to rename the file client.py to zammad_rest_client.py and make the class name ZammadRestClient

mcp-servers/zammad/src/zammad_mcp/client.py
```diff
@@ -0,0 +1,185 @@
...
+            try:
+                data = r.json()
+            except json.JSONDecodeError as e:
+                raise ValueError(
+                    "Zammad REST returned non-JSON (check ZAMMAD_URL / reverse proxy)."
+                ) from e
+            return dict(data)
+
+
+def get_zammad_rest_client(http_token: str | None = None) -> ZammadClient:
```
mhdawson: I think this could be private to the file as it's not needed externally any more.

.github/workflows/pr-checks.yml
```diff
@@ -13,7 +13,9 @@ env:
   ASSET_MANAGER_DIR: asset-manager
   # uv version must match Makefile UV_VERSION (currently 0.8.9)
   # Update this when UV_VERSION in Makefile changes
-  UV_VERSION: "0.8.9"    
+  UV_VERSION: "0.8.9"
+  # Makefile aggregate deps/reinstall skip Zammad MCP unless this is set (default off locally).
+  ZAMMAD_MCP_ENABLED: "1"
```
mhdawson: ❯ does it make sense for the lint and test to run if the dependencies are not set ?                                               
                                                                                    
● No, it doesn't. If deps-mcp-zammad hasn't been run, lint-mcp-zammad and test-mcp-zammad will fail because the virtual           
  environment won't exist. The flag guards deps-all and reinstall-all but not the targets that depend on those dependencies being
  present, which is inconsistent.                                                                                                 
                                                                                                                                  
  Either the lint and test aggregate targets should also respect ZAMMAD_MCP_ENABLED, or the default should be changed to 1 since  
  the zammad MCP is now a first-class part of the project.

mhdawson: Yes, it would make more sense. Currently the flag is inconsistently applied — it gates dependency installation but not linting, 
  testing, or image building/pushing. A developer setting ZAMMAD_MCP_ENABLED=0 to opt out of Zammad would still have:
                                                                                                                                  
  - lint-mcp-zammad running in lint-mypy-per-directory (and likely failing without deps)                                          
  - test-mcp-zammad running in test-all (same problem)                                                                            
  - build-mcp-zammad-image included in build-all-images                                                                           
  - push-mcp-zammad-image, retag-mcp-zammad-image always available                                                                
                                                                                                                                  
  If the flag is meant to indicate "I'm not using Zammad", it should consistently gate all of those targets the same way it       
  currently gates deps-all and reinstall-all. That would make ZAMMAD_MCP_ENABLED=0 a true opt-out rather than just a partial one. 



**Since this is a bigger change lets remove the addition of ZAMMAD_MCP_ENABLED from this PR and then create  JIRA ticket to cover more fully disabling building/linting/testing the artifacts when Zammad is not being used.**

helm/values.yaml
```diff
@@ -274,16 +274,49 @@ mcp-servers:
...
+        tag: "0.0.12"
       port: 8000
       targetPort: 8000
-      transport: http
-      imagePullPolicy: IfNotPresent
+      transport: streamable-http
+      imagePullPolicy: Always
+      extraContainers:
+        - name: basher-zammad-mcp
+          image: ghcr.io/basher83/zammad-mcp:latest
```
mhdawson: I thnk we should pin this to a specific version so that it does not change and break the quickstart. We can update as needed for new features etc.

Prudhvivuda: okay. pinned it to a specific verison.

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: made it private

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: done

mcp-servers/zammad/src/zammad_mcp/client.py
Prudhvivuda: makes sense

mcp-servers/zammad/src/zammad_mcp/settings.py
Prudhvivuda: done

docs/CLAUDE.md
```diff
@@ -23,7 +23,7 @@ This is a self-service agent blueprint implementing a complete AI agent manageme
 All development operations use the root Makefile:
 
 ```bash
-# Install dependencies for all services
+# Install dependencies for all services (add ZAMMAD_MCP_ENABLED=1 to include Zammad MCP, or make deps-mcp-zammad)
```
mhdawson:
```suggestion
# Install dependencies for all services
```

docs/CONTRIBUTING.md
```diff
@@ -86,13 +86,16 @@ Ensure you have the required tools installed:
    make deps-all
    ```
 
+   By default, `deps-all` does **not** install the Zammad MCP tree (`mcp-servers/zammad`). For a full install (same as CI), use `make deps-all ZAMMAD_MCP_ENABLED=1`, or run `make deps-mcp-zammad` when working on ticketing.
```
mhdawson:
```suggestion
```

mcp-servers/zammad/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## What this process does
+
+- **`ZAMMAD_URL`** + **`ZAMMAD_HTTP_TOKEN`** on this pod: ticket/customer identity is checked with **Basher** (`zammad_get_ticket` and, when needed, `zammad_search_users`). **REST** **`GET /api/v1/users/{id}`** runs only for **`send_to_manager_review`**, to read custom user fields (e.g. manager email) that Basher does not expose on the ticket payload.
+- **`ZAMMAD_BASHER_MCP_URL`**: all other Zammad work runs as MCP tool calls to the **Basher** sidecar ([basher83/Zammad-MCP](https://github.com/basher83/Zammad-MCP)) on the same pod—ticket read, updates, tags, articles, user search.
+
+Helm: **`helm/values.yaml`** (`zammad-mcp`, `extraContainers`) and **`helm/templates/_env-helpers.tpl`**. Basher’s container image is **AGPL-3.0**; confirm license compliance before production.
+
+Configuration is loaded once in **`zammad_mcp/settings.py`** via **`load_zammad_mcp_settings()``** into **`ZAMMAD_MCP_SETTINGS`**. **`ZAMMAD_URL`** and **`ZAMMAD_HTTP_TOKEN`** must both be set in the environment at process start (missing values raise **`ValueError`**). Shared HTTP/tracing helpers live in **`mcp-servers/mcp-common`**.
```
mhdawson:
```suggestion
```

shared-clients/README.md
```diff
@@ -152,7 +152,7 @@ async def process_streaming_response(response_stream):
 This package is designed to be used as a local dependency within the self-service agent blueprint project.
 
 ```bash
-# From the project root, install all dependencies
+# From the project root; set ZAMMAD_MCP_ENABLED=1 before make deps-all if you need Zammad MCP deps
```
mhdawson:
```suggestion
# From the project root, install all dependencies
```

.gitignore
```diff
@@ -211,7 +211,7 @@ helm/charts/*.tgz
 helm/zammad/charts/*.tgz
 helm/tmpcharts*
 evaluations/.deepeval/*
-evaluations/results/deep_eval_results/*.json
+# Per-flow eval outputs: see evaluations/.gitignore (results/**/…)
```
mhdawson:
```suggestion
evaluations/results/deep_eval_results/*.json
evaluations/results/ticket_laptop_refresh/*.json
```

docs/CONTRIBUTING.md
```diff
@@ -102,6 +105,8 @@ Ensure you have the required tools installed:
    make test-all
    ```
 
+   `test-all` includes the Zammad MCP package; install its deps first (`ZAMMAD_MCP_ENABLED=1` with `deps-all`, or `make deps-mcp-zammad`).
```
mhdawson:
```suggestion
   `test-all` includes the Zammad MCP package; install its deps first 
```

Makefile
```diff
@@ -246,8 +247,19 @@ ZAMMAD_CREDENTIALS_SECRET ?= $(MAIN_CHART_NAME)-zammad-credentials
 ZAMMAD_ADMIN_EMAIL ?= admin@zammad.local
 ZAMMAD_ADMIN_PASSWORD ?= ZammadR0cks!
 ZAMMAD_AUTOWIZARD_TOKEN ?= ssa-zammad-autowizard-9f3a2b1c
+
+# Include mcp-servers/zammad in aggregate deps-all / reinstall-all when 1, true, or yes (default: off).
+# CI sets ZAMMAD_MCP_ENABLED=1 so test-all / lint-mcp-zammad have deps; use the same locally for full tree, or run deps-mcp-zammad manually.
+ZAMMAD_MCP_ENABLED ?= 0
+ZAMMAD_MCP_IN_AGG_TARGETS := $(filter 1 true yes,$(ZAMMAD_MCP_ENABLED))
+ZAMMAD_MCP_DEPS_ALL_EXTRA := $(if $(ZAMMAD_MCP_IN_AGG_TARGETS),deps-mcp-zammad)
+ZAMMAD_MCP_REINSTALL_ALL_EXTRA := $(if $(ZAMMAD_MCP_IN_AGG_TARGETS),reinstall-mcp-zammad)
```
mhdawson: I think these should be removed along with the other changes, but the deps-mcp-zammad and reinstall-mcp-zammad may need to be integrated back into the list.

Makefile
```diff
@@ -296,11 +309,12 @@ help:
 	@echo "  verify-triggers                     - Verify all expected Knative triggers are deployed"
 	@echo ""
 	@echo "Dependency Commands (local dev):"
-	@echo "  deps-all                            - Install dependencies for all projects"
+	@echo "  deps-all                            - Install dependencies for all projects (Zammad MCP: set ZAMMAD_MCP_ENABLED=1 to include)"
```
mhdawson:
```suggestion
	@echo "  deps-all                            - Install dependencies for all projects"
```

Makefile
```diff
@@ -312,11 +326,12 @@ help:
 	@echo "  deps-evaluations                     - Install dependencies for evaluations"
 	@echo ""
 	@echo "Reinstall Commands:"
-	@echo "  reinstall-all                       - Force reinstall dependencies for all projects (uv sync --reinstall)"
+	@echo "  reinstall-all                       - Force reinstall dependencies for all projects (Zammad MCP: set ZAMMAD_MCP_ENABLED=1 to include)"
```
mhdawson:
```suggestion
	@echo "  reinstall-all                       - Force reinstall dependencies for all projects"
```

Makefile
```diff
@@ -1468,6 +1537,7 @@ helm-depend:
 		echo "Helm dependencies present (Chart.lock), skipping update"; \
 	else \
 		echo "Updating Helm dependencies"; \
+		helm repo update 2>/dev/null || true; \
```
mhdawson: Why was this needed?

Makefile
Prudhvivuda: removed

mhdawson (approved):
LGTM

mhdawson:
@tchughesiv could you take a look I think your comment was addressed

Makefile
```diff
@@ -1350,6 +1406,12 @@ test-integration-dispatcher:
 test-mcp-snow:
 	$(call run_pytest,snow MCP,mcp-servers/snow)
 
+.PHONY: test-mcp-zammad
+test-mcp-zammad:
+	@echo "Running Zammad MCP tests..."
+	cd mcp-servers/zammad && uv run python -m pytest tests/
+	@echo "Zammad MCP tests completed successfully!"
```
tchughesiv: let's match the existing test approach
```suggestion
test-mcp-snow:
	$(call run_pytest,snow MCP,mcp-servers/snow)

.PHONY: test-mcp-zammad
test-mcp-zammad:
	$(call run_pytest,zammad MCP,mcp-servers/zammad)
```

docs/CONTRIBUTING.md
```diff
@@ -102,6 +104,8 @@ Ensure you have the required tools installed:
    make test-all
    ```
 
+   `test-all` includes the Zammad MCP package; install its deps first 
+
```
tchughesiv: unecessary, zammad mcp should be no different than any of the others... no need it call it out specifically.

tchughesiv (changes requested):

evaluations/.gitignore
```diff
@@ -1,3 +1,12 @@
 conversations_config/default_context/*
-results/conversation_results/*
```
tchughesiv, line 2: shouldn't we keep this one?

mcp-servers/snow/src/snow/server.py
tchughesiv: let's fix this, we shouldn't ignore errors in actual code... tests are ok where necessary

mcp-servers/snow/src/snow/server.py
tchughesiv: same

mcp-servers/zammad/src/zammad_mcp/basher_client.py
```diff
@@ -0,0 +1,166 @@
...
+import httpx
+from mcp import ClientSession
+from mcp.client.streamable_http import streamable_http_client
+from mcp.types import CallToolResult, TextContent
+from shared_models import configure_logging
+from zammad_mcp.settings import ZAMMAD_MCP_SETTINGS
+
+logger = configure_logging("zammad-mcp-basher")
+
+_executor = ThreadPoolExecutor(max_workers=8, thread_name_prefix="zammad-basher-mcp")
```
tchughesiv: `max_workers` should this be configurable? what do we do in other mcp servers?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+            "Expected format: {email}-{ticket_id} (Zammad internal ticket id)."
+        )
+    email, ticket_id = parse_email_and_ticket_id(raw)
+    cust_uid = assert_ticket_customer_matches_basher(ticket_id, email)
+    return email, ticket_id, cust_uid
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
```
tchughesiv: let's not ignore issues
```suggestion
@trace_mcp_tool()
```

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+    parts = [f"Ticket {ticket_id} tagged as {tag!r}"]
+    if owner:
+        parts.append(f"owner set to {owner!r}")
+    return ", ".join(parts) + "."
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
```
tchughesiv: same

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+    """Close the ticket."""
+    _, ticket_id, _ = _authorize_ticket(ctx)
+    state = zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.state_closed
+    call_basher_tool("zammad_update_ticket", {"ticket_id": ticket_id, "state": state})
+    return f"Ticket {ticket_id} moved to state {state!r}."
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
```
tchughesiv: same

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+    if update:
+        call_basher_tool("zammad_update_ticket", {"ticket_id": ticket_id, **update})
+
+    gmsg = f", group {gname!r}" if gstrip else ""
+    return f"Ticket {ticket_id} tagged {tag!r}{gmsg}."
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
```
tchughesiv: same

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+    call_basher_tool(
+        "zammad_update_ticket",
+        {"ticket_id": ticket_id, "owner": manager},
+    )
+    return f"Ticket {ticket_id} assigned to {manager!r}, tag {tag!r}."
+
+
+@mcp.tool()
+@_handle_tool_errors
+@trace_mcp_tool()  # type: ignore[untyped-decorator]
```
tchughesiv: same

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+    host=zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.mcp_listen_host,
+    port=zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.mcp_listen_port,
+    stateless_http=(
+        zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.mcp_transport == "streamable-http"
+    ),
+    lifespan=lifespan,
+)
+
+
+@mcp.custom_route("/health", methods=["GET"])  # type: ignore
```
tchughesiv: same

tchughesiv (changes requested):
please don't rebase/squash as you fix these things

mcp-servers/zammad/src/zammad_mcp/client.py
tchughesiv: older setups?

tchughesiv:
@Prudhvivuda make sure you are fully vetting these changes in a cluster

mcp-servers/zammad/src/zammad_mcp/basher_client.py
Prudhvivuda: Okay, making it configurable. In this repo, only the Zammad wrapper uses a ThreadPoolExecutor: it bridges sync FastMCP tool handlers to async Basher MCP calls (asyncio.run per task). Snow does not use that pattern.

evaluations/.gitignore
Prudhvivuda: sure

mcp-servers/zammad/README.md
```diff
@@ -26,6 +26,7 @@ Example: `alice@company.com-42`
 |----------|---------|
 | `ZAMMAD_URL` | Zammad web origin for REST (no `/api/v1` in the value; paths are added in code). Helm secret key **`zammad-url`**. |
 | `ZAMMAD_BASHER_MCP_URL` | Basher MCP URL. Defaults to **`http://127.0.0.1:8001/mcp`** (sidecar on the same pod); set only if you use another host or path. |
+| `ZAMMAD_BASHER_MCP_MAX_WORKERS` | Size of the thread pool used for synchronous Basher MCP calls (default **`8`**; must be an integer, clamped to **`1`**–**`128`**). |
```
tchughesiv, line 29: why was 8 chosen as the default? any impact on memory/cpu usage ?

Prudhvivuda: Well, there is no specific reason and not derived from any benchmarks. AI gave me that number saying it's a reasonable starting default. We can adjust that to any other number if we want. Below is the reason it generated.

Memory / CPU impact

- Threads: Each worker is an OS thread. In practice you pay per-thread stack memory (often on the order of ~1 MiB per thread on Linux, platform-dependent), not “8× full Python interpreters.”
- CPU: CPython’s GIL limits true parallel CPU work across threads, but these Basher calls are mostly I/O (waiting on HTTP). Extra threads mainly help overlap waiting, not burn 8× CPU when idle.
- Load: Under heavy concurrent MCP traffic, more workers can mean more simultaneous Basher connections and more in-flight work (so CPU, file descriptors, and downstream Basher can see more pressure). ZAMMAD_BASHER_MCP_MAX_WORKERS is there so ops can lower it on tight pods or raise it if profiling shows queueing.

helm/values.yaml
```diff
@@ -274,16 +274,49 @@ mcp-servers:
       replicas: 1
       deploymentMode: deployment
       image:
-        repository: ghcr.io/basher83/zammad-mcp
-        tag: "1.1.0"
+        repository: quay.io/rh-ai-quickstart/self-service-agent-zammad-mcp
+        tag: "0.0.12"
```
tchughesiv: shouldn't this be `0.0.13` ?
I intend to make this more dynamic with https://github.com/rh-ai-quickstart/it-self-service-agent/pull/359

tchughesiv (changes requested):

helm/values.yaml
Prudhvivuda: thanks for this. bumped the tag.

mhdawson: Maybe it is that we need to update the workflow that bumps that version to update the new containers added for the ticket quickstart as well?

This is what is already beeing bumped by the workflow https://github.com/rh-ai-quickstart/it-self-service-agent/pull/350/changes

mhdawson: I see that you are fixing what I mentioned in https://github.com/rh-ai-quickstart/it-self-service-agent/pull/359https://github.com/rh-ai-quickstart/it-self-service-agent/pull/359

tchughesiv (approved):
lgtm

PR #347, opened 2026-03-27
feat: add phase 1 ticket agents

mhdawson:
- fix default agent config so that it can be set with DEFAULT_AGENT_ID_ARG environment variable when deploying.
- add phase 1 ticket agents (review and laptop-refresh) in big,small and scout variants. In phase 1 agents just state what they will do instead of using mcp servers to change ticket state. Evals now pass for multiple runs of 20 for each variant.
- tweak evals to generate conversations correctly for ticket flow.
- tweaks to evals to add options needed to drive conversation in way needed by ticket flow. Should have no effect on default flow.
- add additional predefined conversations to cover additional scenarios
- fix mock data so John Doe is not eligible which was the case before march 20 2026.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -482,6 +482,10 @@ help:
 	@echo "      LG_PROMPT_EMAIL_UPDATE          - Override email-update agent"
 	@echo "    Usage: make helm-install-test LG_PROMPT_LAPTOP_REFRESH=config/lg-prompts/custom.yaml"
 	@echo ""
+	@echo "  Routing Agent Configuration:"
+	@echo "    DEFAULT_AGENT_ID                  - Override the default routing agent (default: routing-agent)"
+	@echo "    Usage: make helm-install-test DEFAULT_AGENT_ID=my-custom-agent"
+	@echo ""
 	@echo "  Evaluation Configuration:"
 	@echo "    VALIDATE_FULL_LAPTOP_DETAILS    - Enable full laptop details validation (default: true)"
 	@echo "                                        Set to 'false' to disable: VALIDATE_FULL_LAPTOP_DETAILS=false"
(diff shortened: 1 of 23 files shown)


Makefile
```diff
@@ -1558,14 +1562,16 @@ endef
...
 
 .PHONY: helm-install-test
 helm-install-test: namespace helm-depend
 	$(call helm_install_common,"with mock eventing service - testing/CI",\
 		-f helm/values-test.yaml \
 		--set requestManagement.knative.mockEventing.enabled=true \
 		--set testIntegrationEnabled=true \
-		$(PROMPT_OVERRIDES),\
+		$(PROMPT_OVERRIDES) \
+		$(DEFAULT_AGENT_ID_ARG),\
```
tchughesiv: this seems like it may make more sense in the `helm_install_common` method, so all helm installs get the change.

tchughesiv (changes requested):

Makefile
mhdawson: @tchughesiv good catch updated.

Makefile
```diff
@@ -1818,7 +1824,8 @@ helm-export-demo: helm-depend
 		-f helm/values-test.yaml \
 		-f helm/values-demo.yaml \
 		$(helm_demo_email_args) \
-		$(PROMPT_OVERRIDES))
+		$(PROMPT_OVERRIDES) \
+		$(DEFAULT_AGENT_ID_ARG))
```
tchughesiv: similarly, should this be set in the `helm_export` method instead?

mhdawson: ok done

mhdawson: @tchughesiv

tchughesiv (approved):
lgtm

PR #352, opened 2026-04-01
fix(helm): pin integration-dispatcher to one Uvicorn worker

tchughesiv:
Integration-dispatcher runs IMAP + outbox in **lifespan**; multiple Uvicorn workers duplicate that work and stress CPU/memory/DB pools. This PR **always sets `UVICORN_WORKERS=1`** in `_service-deployment.tpl` for that service and removes the Helm value. **request-manager** and **agent-service** still use `uvicornWorkers` from values.

## Symptoms
- `Child process [4] died` / several `Started server process [N]`
- Startup probe: `connection refused` on `:8080`; restarts; exit **137**
- CrashLoopBackOff on a new ReplicaSet while an older pod still looked healthy

## Changes
- Template: fixed `UVICORN_WORKERS` for integration-dispatcher only  
- `values.yaml` / `values-demo.yaml`: removed `integrationDispatcher.uvicornWorkers`  
- `guides/PERFORMANCE_SCALING_GUIDE.md`: scale integration-dispatcher by **replicas**, not workers

## Verify
`helm template …` for integration-dispatcher → `UVICORN_WORKERS: "1"`

### DIFF
--- a/guides/PERFORMANCE_SCALING_GUIDE.md
+++ b/guides/PERFORMANCE_SCALING_GUIDE.md
@@ -42,7 +42,7 @@ Asyncio is a great fit for this quickstart as we've seen that most of the time a
 
 ### Quickstart components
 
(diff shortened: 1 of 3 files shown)


mhdawson:
@tchughesiv I think we've decide we won't land this right. Instead creating a PR to change the uvicorn worker to 2, right?

PR #353, opened 2026-04-02
feat: load all test users into zammad

mhdawson:
- used new init container to load users into zammad as existing
      method could not handle adding to groups etc.
    - loads all users from demo data
    - creates manager users
    - sets first manager user email on first 5 demo users and second on the rest
    - creates ticket_handler1 and ticket_handler2 users which are in the
      human_managed_tickets group
    - creates agent.latop-specialist user for the laptop refresh specialist agent


mhdawson:
Converting to draft as it includes commits from my phase 1 ticket agents PR. Will want to land after that one.

mhdawson:
other PR landed, rebased so ready for review/landing

Makefile
```diff
@@ -2078,24 +2111,25 @@ deploy-zammad: namespace
...
+		echo "  extraEnv:" >> $$TMPFQDN; \
+		echo "    - name: ZAMMAD_FQDN" >> $$TMPFQDN; \
+		echo "      value: \"$(ZAMMAD_FQDN)\"" >> $$TMPFQDN; \
+		echo "    - name: ZAMMAD_HTTP_TYPE" >> $$TMPFQDN; \
+		echo "      value: \"https\"" >> $$TMPFQDN; \
 		ZAMMAD_ARGS="$$ZAMMAD_ARGS -f $$TMPFQDN"; \
 	fi; \
-	helm upgrade --install zammad zammad/zammad \
-		--version $(ZAMMAD_CHART_VERSION) \
+	helm upgrade --install zammad-stack helm/zammad/ \
```
tchughesiv: why change the name from `zammad` to `zammad-stack`? "stack" is implied and makes object names longer across board.

tchughesiv (changes requested):
if we're creating a separate zammad template dir, we should probably also move the existing `helm/templates/zammad-embed-*.yaml` files.

Makefile
mhdawson: good point, fixed.

mhdawson:
@tchughesiv moved templates into 

> if we're creating a separate zammad template dir, we should probably also move the existing `helm/templates/zammad-embed-*.yaml` files.

@tchughesiv done, took a few related changes but makes the zammod deploy more self contained so good overall.

Makefile
```diff
@@ -2188,26 +2206,6 @@ zammad-bootstrap-token: namespace zammad-trigger-autowizard
 	kubectl rollout status deployment/mcp-zammad-mcp -n $(NAMESPACE) --timeout=2m; \
 	echo "✅ Zammad token created and MCP restarted."
 
-# Update Zammad API token and restart MCP. Run after creating token in Zammad Admin → Token Access.
-# Update embed page with actual Zammad URL (fixes placeholder when embed shows YOUR-ZAMMAD-URL)
-.PHONY: zammad-update-embed-url
-zammad-update-embed-url: namespace
```
tchughesiv, line 2194: how is this handled now? were you able to test the embed page example successfully?

tchughesiv (commented):
looking better

Makefile
mhdawson: yes, I deployed the URL was shown and I viewed the page.

mhdawson: It is now a separate target which is deployed after zammad is fully deployed, this makes sure we have what's needed for the deployment without having to do a restart.

There may still be some tweaking to optimize/reduce restarts etc. but this made sense for the current refactor.

tchughesiv (approved):
lgtm

PR #356, opened 2026-04-13
fix: Update README,EVALUATIONS_GUIDE, and uv files with max supported Python version (3.13)

sgratch:
```
 - Add notes for both README.md, guides/EVALUATIONS GUIDE.md guides regarding
    Python supported versions 3.12, 3.13 only since 'grpcio' library
    doesn't have wheels for latest Python 3.14 yet.
    
 - Update pyproject.toml, uv.lock files with those supported Python versions.

Signed-off-by: Sharon Gratch <sgratch@redhat.com>
```


tchughesiv (commented):
thanks for the contrib! i wonder though if, rather than add python version specific notes... if instead we should harden up our dev requirements in the README? i think we currently state we support python 3.12+ ... but maybe we just change it to 3.12 ... thoughts @mhdawson  ?

sgratch:
> thanks for the contrib! i wonder though if, rather than add python version specific notes... if instead we should harden up our dev requirements in the README? i think we currently state we support python 3.12+ ... but maybe we just change it to 3.12 ... 

@tchughesiv  @mhdawson  yes, I  thought about it but there are 2 concerns:
-  3.13 is also supported in that context so anyway worth changing to 3.12 - 3.13 
- Instead of asking the user to downgrade his system Python for it, in case 3.14 is installed, I thought it's better to let him know that the  workaround is also ok  (i.e. `uv sync --python 3.13`) for running the evaluations.

I can move the note to the dev requirements section and not next to the uv sync examples though..
WDYT?

mhdawson:
Mabye the middle ground is something like

**Note:** If your system Python is 3.14, or higher run `uv sync --python 3.13` instead of `uv sync` since the dependent `grpcio`  
library doesn't have pre-compiled binary packages the most recent versions of Python yet.

It's a bit less tied to which later python version you are running. @tchughesiv what do you think?

tchughesiv:
I think that makes sense... I also like the idea of changing the supported versions in the readme.

mhdawson:
@sgratch can you update based on my suggestion and then we can trigger the full CI run to start.

sgratch:
> can you update based on my suggestion and then we can trigger the full CI run to start.

@mhdawson @tchughesiv  fixed, please review again

mhdawson (approved):
LGTM

tchughesiv (changes requested):
please change the 3.12+ requirements doc in README to be 3.12-3.13

tchughesiv (approved):
Lgtm

mhdawson (approved):
LGTM

PR #358, opened 2026-04-16
update laptop option prompts

gurvinder:
update laptop option classification to classify user responses based on overall intent rather than individual words in the lg-prompt-small and lg-prompt-small-scout prompt configs

[make_test_long_resp_integration_request_mgr.log](https://github.com/user-attachments/files/26799116/make_test_long_resp_integration_request_mgr.log)

### DIFF
--- a/agent-service/config/lg-prompts/lg-prompt-small-scout.yaml
+++ b/agent-service/config/lg-prompts/lg-prompt-small-scout.yaml
@@ -297,6 +297,8 @@ states:
 
       The user was asked if they want to proceed with reviewing laptop options. They responded: "{user_input}"
 
+      IMPORTANT: Classify based on the user's overall intent, not individual words in isolation. If the response expresses a desire to continue with or see laptop options, classify as YES.
+
       Classify their response as one of the following:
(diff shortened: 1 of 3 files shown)


tchughesiv (approved):
lgtm

mhdawson:
Overall looks good, but going to run a few longer night CI runs as an additional sanity check.

mhdawson:
Run of 20 small scout -https://github.com/rh-ai-quickstart/it-self-service-agent/actions/runs/24567245066 - PASSED

mhdawson:
70b - Big prompt - https://github.com/rh-ai-quickstart/it-self-service-agent/actions/runs/24574732797

Unforutnately for first run on this, the new test failed. Re-running to see if it repeats.

It passed on the second run which unfotunately means its intermittent. 

@gurvinder could you run a bunch of times with 70 and the big prompt so we can understand how flaky the test will be? If its more than something like 1/100 I'm not sure we'll want to add the test for now.

mhdawson (changes requested):
marking as request changes until we understand how of the test fails with the large prompt.

mhdawson:
From the runs that Gurvinder did it seems like this should be flaky in less than 1/10 runs, less flaky than another failure we are seeing in the nightly runs so I'm going to go ahead and land.

mhdawson (approved):
LGTM

PR #359, opened 2026-04-20
chore: centralize version bumps and align Zammad bootstrap with ticketing Helm

tchughesiv:
- Add `scripts/bump-release.sh` / `bump_release.py` + manifest to bump `BASE_VERSION`, Chart `appVersion`, `helm/values.yaml` release image anchor, and Zammad bootstrap `imageTag`; skip unchanged files and fix drift when Makefile already matches.
- Zammad bootstrap: registry/repo/tag only in values/templates; Zammad MCP image tag aliases `releaseImageTag`.
- Update dev→main and dev version-bump workflows for YAML anchors/aliases; version-bump workflow runs the bump script instead of `sed`.

### DIFF
--- a/.github/workflows/create-dev-to-main-pr.yaml
+++ b/.github/workflows/create-dev-to-main-pr.yaml
@@ -139,13 +139,11 @@ jobs:
           # Use the target commit that was already determined
           DEV_REF="${{ steps.check_commits.outputs.target_commit }}"
 
-          # Extract BASE_VERSION from Makefile in dev
+          # Align Makefile BASE_VERSION with every path/scripts/bump-release.manifest.json (Chart, Helm tags, Zammad bootstrap)
+          python3 scripts/bump_release.py --verify --git-ref "$DEV_REF"
+
           DEV_MAKEFILE_VERSION=$(git show "$DEV_REF:Makefile" | grep '^BASE_VERSION' | head -1 | sed 's/BASE_VERSION.*= *//')
 
-          # Extract appVersion from helm/Chart.yaml in dev
-          DEV_HELM_VERSION=$(git show "$DEV_REF:helm/Chart.yaml" | grep '^appVersion:' | sed 's/appVersion: *//')
-
-          # Extract BASE_VERSION from Makefile in main (fallback to VERSION if not found)
           MAIN_MAKEFILE_VERSION=$(git show origin/main:Makefile | grep '^BASE_VERSION' | head -1 | sed 's/BASE_VERSION.*= *//')
           MAIN_VERSION_VAR="BASE_VERSION"
           if [ -z "$MAIN_MAKEFILE_VERSION" ]; then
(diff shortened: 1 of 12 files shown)


.github/workflows/create-dev-to-main-pr.yaml
```diff
@@ -197,9 +201,8 @@ jobs:
             echo ""
             echo "❌ helm/values.yaml image.tag ($DEV_IMAGE_TAG) does not match Makefile BASE_VERSION ($DEV_MAKEFILE_VERSION)"
             echo ""
-            echo "To fix:"
-            echo "  1. Update helm/values.yaml image.tag to match Makefile BASE_VERSION ($DEV_MAKEFILE_VERSION)"
-            echo "  2. Commit and push the change to dev"
+            echo "To fix: ./scripts/bump-release.sh $DEV_MAKEFILE_VERSION"
```
mhdawson: Did you mean to leave the "To fix" in this ?

tchughesiv: yes, this is displayed to user in actions logs when there's a version mismatch that needs fixing.

tchughesiv: its just replacing the prior message that displayed something similar

mhdawson: k, got it.

helm/zammad/values.yaml
```diff
@@ -9,8 +9,11 @@ externalRoute:
   enabled: false
 
 bootstrap:
-  image: "quay.io/rh-ai-quickstart/self-service-agent-zammad-bootstrap:latest"
-  # Must match autoWizard.config Users[0] and Token in helm/values-ticketing.yaml
+  # imageTag ↔ Makefile BASE_VERSION / scripts/bump-release.sh. Ticketing overrides via helm-install-ticketing overlay.
+  imageRegistry: quay.io/rh-ai-quickstart
+  imageRepository: self-service-agent-zammad-bootstrap
+  imageTag: "0.0.13"
```
mhdawson, line 15: Other places now use  `tag: *releaseImageTag` instead of `0.0.13` directly, should that be used here as well?

tchughesiv: wish we could... variable is limited to same file

.github/workflows/create-dev-version-bump-pr.yml
```diff
@@ -98,8 +94,8 @@ jobs:
           # Create and checkout new branch
           git checkout -b "$BRANCH_NAME"
 
-          # Stage and commit changes
-          git add Makefile helm/Chart.yaml helm/values.yaml
+          # Stage and commit changes (paths must match scripts/bump-release.manifest.json)
+          git add Makefile helm/Chart.yaml helm/values.yaml helm/zammad/values.yaml
```
mhdawson: Not a blocker if its not easy but I wonder if there is any way we can use the paths defined in bump_release.py to generete the list of files to be added as that would close one more opportunity for misisng bumping one of the files.

tchughesiv: done

helm/zammad/values.yaml
mhdawson: k

mhdawson (approved):
LGTM

PR #360, opened 2026-04-22
fix: optimize deployment - step 1

mhdawson:
- deploy both main quickstart and zammad in parallel. Cuts deployment time down to 3-4 minutes
- move towards more helm native deploy for the zammad components
- remove ability to deploy/undeploy zammad on its own


mhdawson:
I named this part 1 as the deploy still has the zammad containers crashing/retrying a lot we should see if we can improve on that.

zammad-bootstrap/bootstrap.py
```diff
@@ -299,6 +301,96 @@ def get_or_create_user(
         print(f"  Created user '{email}' (id={user['id']})")
 
 
+# ---------------------------------------------------------------------------
+# Token creation + Kubernetes secret/deployment update
+# ---------------------------------------------------------------------------
```
tchughesiv, line 308: if we're going to interact with k8s from here, prob best to leverage python client library instead -
https://pypi.org/project/kubernetes/

tchughesiv (changes requested):

zammad-bootstrap/bootstrap.py
mhdawson: Makes sense, thanks for catching that, will update.

mhdawson: @tchughesiv pushed a commit to fix.

tchughesiv (approved):
lgtm

PR #361, opened 2026-04-24
export OTLP to Zammad MCP

Prudhvivuda:
This change ensures distributed tracing works for the Zammad MCP server the same way it already does for the Snow MCP server. The Makefile previously passed OTEL_EXPORTER_OTLP_ENDPOINT only to self-service-agent-snow, so the Zammad MCP process never initialized OpenTelemetry and no `zammad-mcp-server` spans appeared in Jaeger even when ticket mode was deployed correctly.

The Helm install path now sets OTEL_EXPORTER_OTLP_ENDPOINT on `mcp-servers.mcp-servers.zammad-mcp` as well, using `--set-string` so OTLP URLs are passed reliably.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,8 +215,7 @@ helm_generic_args = \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_METRICS_EXPORTER=none,) \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_LOGS_EXPORTER=none,) \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf,) \
-	$(if $(OTEL_EXPORTER_OTLP_ENDPOINT),--set mcp-servers.mcp-servers.self-service-agent-snow.env.OTEL_EXPORTER_OTLP_ENDPOINT="$(OTEL_EXPORTER_OTLP_ENDPOINT)") \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -215,7 +215,8 @@ helm_generic_args = \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_METRICS_EXPORTER=none,) \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_LOGS_EXPORTER=none,) \
 	$(if $(findstring jaeger,$(OTEL_EXPORTER_OTLP_ENDPOINT)),--set-string llama-stack.secrets.OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf,) \
-	$(if $(OTEL_EXPORTER_OTLP_ENDPOINT),--set mcp-servers.mcp-servers.self-service-agent-snow.env.OTEL_EXPORTER_OTLP_ENDPOINT="$(OTEL_EXPORTER_OTLP_ENDPOINT)")
+	$(if $(OTEL_EXPORTER_OTLP_ENDPOINT),--set mcp-servers.mcp-servers.self-service-agent-snow.env.OTEL_EXPORTER_OTLP_ENDPOINT="$(OTEL_EXPORTER_OTLP_ENDPOINT)") \
+	$(if $(OTEL_EXPORTER_OTLP_ENDPOINT),--set mcp-servers.mcp-servers.zammad-mcp.env.OTEL_EXPORTER_OTLP_ENDPOINT="$(OTEL_EXPORTER_OTLP_ENDPOINT)")
```
mhdawson: Should this also be guarded by whether the zammad MCP server is enabled?

Prudhvivuda: Yes. Pushing mcp-servers.mcp-servers.zammad-mcp.env.OTEL_EXPORTER_OTLP_ENDPOINT from helm_generic_args meant every install path (test, demo, prod) was writing Zammad MCP values even though zammad-mcp stays enabled: false there.

Prudhvivuda: Moving it to ticketing args...

That --set was removed from helm_generic_args and added to helm_ticketing_args, which is only used by _helm-install-ticketing-single together with values-ticketing.yaml and --set mcp-servers.mcp-servers.zammad-mcp.enabled=true

mhdawson (approved):
LGTM

PR #362, opened 2026-04-28
feat: add general agent

mhdawson:
- add general agent that will later use knowledge base to answer questions
- route non laptop refresh tickets to general agent instead of escalating
- update metrics for ticket_unrelated flow based on addition of general agent
- update generator for ticket_unrelated flow based on addition of general agent
- add known bad and predefined conversations for ticket_unrelated flow
- update evaluation framework so we can specify multiple flows to evaluate.py
- update short target for ticket flow to run both ticket_laptop_refresh and ticket_unrelated flows


tchughesiv (approved):
lgtm

PR #364, opened 2026-04-30
Landing page with user/passwords for Demo

Prudhvivuda:
Adds a path-based demo login portal on the same host as Zammad (e.g. /demo-portal/), with the chat embed snippet on a separate static page. Routing is handled by OpenShift Routes: Zammad serves /, the embed nginx serves spec.path; nginx uses alias under that prefix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2173,7 +2173,8 @@ undeploy-email-server:
 	@kubectl delete -f test-email-server/test-email-server-greenmail.yaml -n $(NAMESPACE) --ignore-not-found 2>/dev/null || true
 	@echo "✅ Test email server removed successfully!"
 
-# Optional helm/zammad-embed (Route ssa-zammad-embed: same host as ssa-zammad, path /$(ZAMMAD_EMBED_PATH)/). Literal true installs; anything else skips install and runs helm uninstall so a prior embed is removed.
+# Optional demo portal served from the Zammad host at /$(ZAMMAD_EMBED_PATH)/.
+# Set ZAMMAD_EMBED_ENABLED=true to install it; any other value removes it.
 ZAMMAD_EMBED_ENABLED ?= false
 # URL path segment for the demo portal (must match helm --set pathPrefix=/$(ZAMMAD_EMBED_PATH)).
 ZAMMAD_EMBED_PATH ?= demo-portal
--- a/Makefile
+++ b/Makefile
@@ -1780,7 +1780,7 @@ _helm-install-ticketing-print-checklist:
 		echo "     Web UI: https://$$ZAMMAD_ROUTE"; \
 		echo "     API:    https://$$ZAMMAD_ROUTE/api/v1"; \
 		if [ -n "$$ZAMMAD_EMBED_ROUTE" ]; then \
-			echo "     Embed:  https://$$ZAMMAD_EMBED_ROUTE (Zammad chat widget — embed snippet + live preview)"; \
+			echo "     Embed:  https://$$ZAMMAD_EMBED_ROUTE (split-screen demo login + Zammad iframe; /chat-embed.html = chat widget)"; \
 		else \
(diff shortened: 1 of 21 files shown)


Makefile
```diff
@@ -2172,8 +2173,11 @@ undeploy-email-server:
 	@kubectl delete -f test-email-server/test-email-server-greenmail.yaml -n $(NAMESPACE) --ignore-not-found 2>/dev/null || true
 	@echo "✅ Test email server removed successfully!"
 
-# Optional helm/zammad-embed (ssa-zammad-embed Route + snippet page). Literal true installs; anything else skips install and runs helm uninstall so a prior embed is removed.
+# Optional demo portal served from the Zammad host at /$(ZAMMAD_EMBED_PATH)/.
+# Set ZAMMAD_EMBED_ENABLED=true to install it; any other value removes it.
```
mhdawson: Does ZAMMAD_EMBED_ENABLED still make sense since its not a portal page and not the embed ?

helm/zammad-embed/values.yaml
```diff
@@ -3,6 +3,124 @@ enabled: false
 # publicUrl: Zammad base URL for the chat widget script src.
 # If not set, auto-detected from the ssa-zammad OpenShift Route via Helm lookup.
 publicUrl: ""
+
```
mhdawson: I guess same general question does having this tied to the name zammad-embed make sense or should it be a completely different thing.

@tchughesiv what's your take.

Makefile
Prudhvivuda: ZAMMAD_EMBED_ENABLED toggles the whole optional zammad-embed Helm chart not a single page. That chart installs the Route ssa-zammad-embed, nginx, and a ConfigMap that includes both the demo login portal at `/<path>/ `and `/<path>/chat-embed.html` for the chat widget snippet + preview.

So the flag name lines up with the chart/release name (zammad-embed). If we renamed it for clarity we could use something like ZAMMAD_EMBED_CHART_ENABLED or ZAMMAD_DEMO_SITE_ENABLED, but functionally it’s “optional same‑host static bundle under pathPrefix.”

mhdawson: Right I'm not sure having the embed and demo site tied together makes sense, but I think @tchughesiv can better comment on that and what naming would be good.

tchughesiv: i think it's prob ok to tie them together, but maybe we change the var name and it defaults to true now?

Prudhvivuda: Ok. Renamed to ZAMMAD_EMBED_CHART_ENABLED, documented that it toggles the whole zammad-embed chart (demo + chat snippet), and defaulted it to true. Callers that used ZAMMAD_EMBED_ENABLED should switch to ZAMMAD_EMBED_CHART_ENABLED; use `=false` to skip the chart.

mhdawson: @Prudhvivuda , I think it would be better to name it something like ZAMMAD_DEMO_SITE_ENABLED, with that naming flowing through other parts as well (ie embedd would mostly be gone) and then the embedd is just one thing on the demo site

Agree having it enabled by default makes sense.

Prudhvivuda: Renamed the Makefile toggles to ZAMMAD_DEMO_SITE_ENABLED / ZAMMAD_DEMO_SITE_PATH. The Helm release name stays zammad-embed (and Route ssa-zammad-embed) so we don’t churn cluster state / automation.

helm/zammad-embed/templates/_demo-portal.tpl
```diff
@@ -0,0 +1,636 @@
+{{- define "zammad-embed.demoPortalIndex" -}}
```
mhdawson: The file name still says zammad-embed and same with what is being defined here...

Prudhvivuda: Thanks for this. I replaced zammad-embed with helm/zammad-demo-site and renamed the Helm defines to zammad-demo-site.* (e.g. demoSiteIndex in _demo-site.tpl)

mhdawson (approved):
LGTM

PR #365, opened 2026-05-01
feat: add option to integrate with trusty AI safety shields

mhdawson:
- adds deployment for nemo guardrails through trusty AI
- requires OpenShift AI with Trusty AI enabled
- optionally deploys jailbreak nim. This is optional as it requires a GPU and NVIDIA credentials with enterprise entitlement
- introduces new USE_NEMO_GUARDRAILS option which calls nemo guardrails on raw user input message and agent message just before it is returned to user as opposed to all responses API calls
- uses self check to add prompt injection on user messages and proper behaviour check on agent responses.
- includes hard coded output check on "breakfast restaurant" so that we can trigger a blocked output by making a request in which the agent will include breakfast restaurant in the response.
- Original safety shields using llama stack remain so we don't break the it-self-service-quickstart, but I've been told they are being deprecated in llama stack and will be removed in a future version. So after we test with the nemo guardrails deployment we should remove the llama stack guardrails and update the README.md to show the use of the nemo guardrails instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,7 @@ MAIN_CHART_NAME := self-service-agent
 HELM_EXPORT_DIR ?= ansible/helm-export
 TOLERATIONS_TEMPLATE=[{"key":"$(1)","effect":"NoSchedule","operator":"Exists"}]
 INGRESS_PREFIX := ssa
+NEMO_GUARDRAILS_CHART ?= helm/nemo-guardrails
 
 # Slack Configuration - only when ENABLE_SLACK set to true
 ifeq ($(ENABLE_SLACK),true)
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -2230,3 +2231,63 @@ servicenow-bootstrap-create-evaluation-users:
 	@echo "Creating evaluation users..."
 	@cd scripts/servicenow-bootstrap && uv run python -m servicenow_bootstrap.create_evaluation_users
 	@echo "Evaluation users creation completed successfully!"
+
+# NeMo Guardrails deployment targets
+.PHONY: deploy-nemo-guardrails
+deploy-nemo-guardrails: namespace
```
tchughesiv, line 2237: should these new commands be added to `make help` output?

tchughesiv (approved):
might be good to add some documentation? otherwise, lgtm

Makefile
mhdawson: I'll add that along with the documentation as we do the README.md to cover how to use them

PR #366, opened 2026-05-01
Zammad login fix

Prudhvivuda:
**What was going wrong**
- Zammad’s login API expects a small “fingerprint” string that identifies the browser (for device/security logging).
- The full Zammad website sets that when you use it. Our demo page was not sending it, so the first time you used persona quick sign-in (without opening Zammad first), the server could answer with “Need fingerprint param!”.
- Opening “Open Zammad” once made it seem “fixed” because the real Zammad UI had already supplied that fingerprint in your session.

**What we changed (the fix)**
The demo page now creates a stable random fingerprint in the browser (and remembers it for that tab), and sends it on both steps it already used:
- signshow (to get the CSRF token)
- signin (actual login)

So persona sign-in works on the first try, without needing to open Zammad first, including in incognito.


mhdawson (approved):
LGTM

PR #367, opened 2026-05-03
feat: Add optional Zammad ticketing channel/webhook

tchughesiv:
## Optional Zammad ticketing channel

Adds an optional **Zammad**-based ticket channel to the blueprint: customers interact via **tickets** instead of only Slack/web/email. The stack ingests Zammad webhooks, runs the same **agent-service** LangGraph flows with ticket-specific agents and prompts, and posts **customer-visible** replies back through **Zammad REST**. Ticket **mutations** (tags, state, specialist tools) go through **Zammad MCP** / Basher, consistent with the rest of the repo.

### End-to-end flow

1. **Zammad** (subchart + bootstrap) is deployable via **`make helm-install-ticketing`** and optional **`values-ticketing.yaml`** overlay.
2. **integration-dispatcher** exposes **`POST /zammad/webhook`**, verifies **HMAC**, dedupes and orders work per ticket, and forwards normalized requests to **request-manager** with ticket metadata (**`ticket_id`**, **`ticket_title`**, owner fields).
3. **request-manager** treats **Zammad** as a first-class integration: schemas, normalizer, routing into the agent pipeline, and **per-ticket sessions** (DB migration **`002`**, **`IntegrationType.ZAMMAD`**, session rules aligned with ticket-scoped IDs).
4. **agent-service** loads **ticket-review**, **ticket-laptop-refresh**, and **ticket-general** configs and **ticket-routing** / laptop LangGraph prompts for classify → tag/MCP → specialist flows; session handling covers Zammad resume, routing, and specialist behavior.
5. **Outbound**: **`ZammadIntegrationHandler`** posts **`ticket_articles`** using a shared HTTP client; optional **`From`** / fallback behavior is configurable.
6. **mcp-servers/zammad** wraps Basher/Zammad API tools agents call from ticket flows.
7. **Bootstrap / demo**: **`zammad-bootstrap`** seeds users, triggers, webhook wiring, and related chart hooks; demo-site templates align with the ticketing layout where applicable.

### Also in scope

- **Evaluations**: **`ticket_laptop_refresh`** and **`ticket_unrelated`** flows and harness glue.
- **Tests**: dispatcher webhook/handler tests, Zammad MCP server tests, RM normalizer updates, etc.
- **Docs**: **`docs/API_REFERENCE.md`** (webhook), **`docs/CLAUDE.md`**, **`mcp-servers/zammad/README.md`**, **`test/README.md`** for the ticket CLI harness.


zammad-bootstrap/bootstrap.py
```diff
@@ -20,18 +59,38 @@
...
-
-BASE_URL = os.environ["ZAMMAD_BASE_URL"].rstrip("/")
-API_URL = f"{BASE_URL}/api/v1"
-ADMIN_EMAIL = os.environ["ZAMMAD_ADMIN_EMAIL"]
-ADMIN_PASSWORD = os.environ["ZAMMAD_ADMIN_PASSWORD"]
-AUTOWIZARD_TOKEN = os.environ["ZAMMAD_AUTOWIZARD_TOKEN"]
+from mock_employee_data.data import get_employee_data
+
+_BOOTSTRAP_PHASE = os.environ.get("ZAMMAD_BOOTSTRAP_PHASE", "").strip()
+if _BOOTSTRAP_PHASE == "wait-init-job-only":
```
mhdawson: What does having this option do for us ?

tchughesiv: bootstrap and zammad-init jobs essentially fight each other... prolonging zammad startup and causing many more pod restarts (and job crashes). with this change, the bootstrap job let's zammad-init finish first.

mhdawson: Is there a reason we would not want it to always wait for the zammad-init? ie do we need an option?

docs/alternate-journeys/scout-only.md
```diff
@@ -81,6 +81,26 @@ export HF_TOKEN=1234
 Setting LG_PROMPT_LAPTOP_REFRESH as shown above ensures that we are using the small prompt approach
 tuned for Llama-4-Scout-17B-16E.
 
+#### LangGraph defaults vs overrides (two different laptop agents)
```
mhdawson: I don't think we should make any changes here, as this is the scout-only journey alternative to the README.md, so ticketing should not affect this alternative journey.

tchughesiv: changed to 1 line addition

docs/alternate-journeys/scout-only.md
```diff
@@ -71,6 +71,8 @@ export LLM_API_TOKEN=your-api-token
 export LLM_URL=https://your-llm-endpoint/v1
 # Optional: export LLM_MAX_TOKENS=2048   # max output tokens (default); keeps input+output within context
 export LG_PROMPT_LAPTOP_REFRESH=/app/agent-service/config/lg-prompts/lg-prompt-small-scout.yaml
+# Optional — ticketing (Zammad): same Scout tuning for the ticket laptop specialist
```
mhdawson: I really don't want any changes in this journey. It is intended only for people running the it-self-service-agent quickstart so at least to start should have no references to the ticket option. I think can only add confusion at this point versus any value.

We might change that at some point, but we'd to have it fully tested and figure out if/how we mention the ticket option versus is being coverered in the separate speeding up ticket resolution quickstart.

tchughesiv: hmm, ok... but my only counterpoint would be that ticketing is part of it-self-service-agent quickstart now. and in order to run it with scout, that's required setting. i had to figure this out for myself and a user not entering in from the speeding up ticketing wrapper repo won't be aware.

mhdawson: People should not be running ticketing without going through the ticketing quickstart directions. We'll have to document it there.

I still think mixing it into the directions for the it-self-service-agent quickstart will add confusion as I don't want us to have to test/support running it-self-service-agent with ticketing unless they are following the ticketing instructions.

So still think we should remove this.

tchughesiv: ok, i'll remove

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -239,17 +259,30 @@ def send_to_manager_review(ctx: Context[Any, Any], dummy_parameter: str = "") ->
...
+    """
     _, ticket_id, _ = _authorize_ticket(ctx)
-    gname = zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.group_human_managed
+    s = zammad_mcp_settings.ZAMMAD_MCP_SETTINGS
+    tag = s.tag_escalate_human
+    call_basher_tool("zammad_add_ticket_tag", {"ticket_id": ticket_id, "tag": tag})
+
+    gname = s.group_human_managed
     gstrip = gname.strip()
+    owner_mail = s.owner_human_managed.strip()
```
mhdawson: I'm not sure assigning an email here makes sense. It's assigned to a group which could included a number of different users. 

Was this needed for some reason ?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -133,19 +132,26 @@ def mark_as_agent_managed_laptop_refresh(
...
+    tag = s.agent_managed_tag
     call_basher_tool("zammad_add_ticket_tag", {"ticket_id": ticket_id, "tag": tag})
 
-    owner = zammad_mcp_settings.ZAMMAD_MCP_SETTINGS.laptop_specialist_owner
+    owner = s.laptop_specialist_owner
+    state_ip = s.state_in_progress
+    # Basher ``TicketUpdateParams`` accepts ``state``, ``owner``, ``group`` as names/emails — not ``*_id``.
+    payload: dict[str, Any] = {}
     if owner:
-        get_user_id_by_email(owner)
```
mhdawson, line 141: Is this no longer needed because instead of configuring an email, we configure a userid ?

mcp-servers/zammad/src/zammad_mcp/server.py
```diff
@@ -239,17 +259,30 @@ def send_to_manager_review(ctx: Context[Any, Any], dummy_parameter: str = "") ->
 def route_to_human_managed_queue(
     ctx: Context[Any, Any], dummy_parameter: str = ""
 ) -> str:
-    """Route this ticket to the human-managed queue."""
+    """Route this ticket to the human-managed queue (general-agent escalation).
+
+    Matches laptop ``escalate_for_human_review`` semantics: adds the human-escalation tag,
+    then updates group (and optional owner when ``ZAMMAD_OWNER_HUMAN_MANAGED`` is set).
```
mhdawson: I had removed this kind of boiler plate from the functions marked @mcp.tool, as it gets added to the LLM context and could make us run out more quickly on the 70b which is context constrained.

If its readlly needed I think adding outside the sef make more sense, otherwise removing makes sense. Same for any other similar additions to anything marked @mcp.tool() call.

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: Basher **v1.1.0** uses string fields for updates, not IDs:

- [`TicketUpdateParams`](https://github.com/basher83/Zammad-MCP/blob/v1.1.0/mcp_zammad/models.py#L348-L357)
- [`zammad_update_ticket`](https://github.com/basher83/Zammad-MCP/blob/v1.1.0/mcp_zammad/server.py#L1159-L1160) → [`Client.update_ticket`](https://github.com/basher83/Zammad-MCP/blob/v1.1.0/mcp_zammad/client.py#L212-L236)

Our `get_user_id_by_email` is only for `assert_ticket_customer_matches_basher` / `AUTHORITATIVE_USER_ID` authorization (`basher_client.py`), not for `zammad_update_ticket` payloads.

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: its just an optional feature in case we want an owner assigned.

we don’t require an assignee for the human-managed group. `ZAMMAD_OWNER_HUMAN_MANAGED` is optional on purpose: unset (default in our chart) means we only move the ticket to `ZAMMAD_GROUP_HUMAN_MANAGED` and tag it; `owner` **is omitted** from the Basher update so it stays a pooled queue.

if a deployment **does** want a concrete assignee (e.g. a dedicated triage owner), they can opt in via that env; otherwise agree that routing to the group alone is the usual case.

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: i had issues with owner assignment working and (i thought) this fixed it. was it working for you previously?

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: good catch. i've reverted to your original definitions/descrips. its worth noting though that `get_employee_laptop_info` has a pretty long one. just fyi in case we want to shorten in future.

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: Got it now... we were only using `get_user_id_by_email` before `zammad_update_ticket` as an existence check for the owner user.

If we kept that check, might we want an explicit policy when it fails (e.g. drop `owner` but still apply `state`, vs fail the whole update). Without that, isn't the check redundant with `zammad_update_ticket` and we might as well save the round-trip and rely on Basher/Zammad to reject a bad owner?

Thoughts?

agent-service/src/agent_service/session_manager.py
```diff
@@ -19,6 +19,37 @@
...
 logging.getLogger("langgraph").setLevel(logging.WARNING)
 
+# Zammad ticket routing:
+# ``sticky_routing_agent`` is persisted when a ticket specialist is first assigned (laptop vs
+# general). Product policy: that decision is final for this ticket — LangGraph must not steer the
+# session to another specialist or back to the routing agent; changing track requires a new Zammad
+# ticket / RM session. Enforcement lives in ``_zammad_apply_sticky_ticket_track`` and specialist
+# lock / no-router handling below.
+STICKY_AGENT_LAPTOP = "ticket-laptop-refresh"
+STICKY_AGENT_GENERAL = "ticket-general-support"
```
mhdawson: This seems hard coded, duplicating values that must match elsewhere?

tchughesiv: true... should we load them at startup from the configs? or maybe make it configurable via env var/helm ?

evaluations/get_deepeval_metrics.py
```diff
@@ -168,21 +168,22 @@ def get_metrics(
             evaluation_steps=[
```
mhdawson, line 168: These metrics are not used for the zammad flows only the original it-self-service-agent flows. 

I don't think we should be making any changes to this file.

The metrics for the two flows used to test the zammad case are in the 

* evaluations/flows/ticket_laptop_refresh/metrics.py and 
* evaluations/flows/ticket_undrelated/metrics.py

Files

helm/values-ticketing.yaml
```diff
@@ -82,14 +175,14 @@ ticketingZammad:
             adaptSecurityContext: force
 
 agent:
-  defaultAgentId: ticket-review-agent
+  defaultAgentId: "routing-agent"
+  zammadDefaultAgentId: "ticket-review-agent"
```
mhdawson: likely related to the earlier discussion of having brought fixed knowledge of agents into the session manager.

helm/values.yaml
```diff
@@ -288,6 +288,9 @@ mcp-servers:
       transport: streamable-http
       imagePullPolicy: Always
       extraContainers:
+        # basher83/Zammad-MCP ``TicketUpdateParams`` (through at least 1.1.0 / main) only allow
+        # ``state``, ``owner``, ``group`` as human-readable names — not ``state_id`` / ``owner_id`` / ``group_id``.
+        # The blueprint wrapper sends those string fields. Pin a digest or bump after upstream adds id fields.
```
mhdawson:
```suggestion
```

integration-dispatcher/src/integration_dispatcher/integrations/defaults.py
```diff
@@ -424,6 +424,13 @@ def _load_default_integrations(self) -> Dict[str, Dict[str, Any]]:
                 "retry_delay_seconds": 10,
                 "config": {},
             },
+            "ZAMMAD": {
+                "enabled": False,  # Updated by health check (ZAMMAD_URL + ZAMMAD_HTTP_TOKEN)
```
mhdawson:
```suggestion
                "enabled": False,  
```

integration-dispatcher/src/integration_dispatcher/integrations/zammad.py
```diff
@@ -0,0 +1,300 @@
...
+                integration_context=ic,
+                ticket_id=ticket_id,
+            )
+            if on_behalf_of:
+                # Zammad warns X-On-Behalf-Of is deprecated; use From.
+                headers["From"] = on_behalf_of
+            response = await client.post(
+                "/ticket_articles", json=payload, headers=headers
+            )
+            if (
```
mhdawson, line 268: I'm wondering when we hit this case ?

integration-dispatcher/src/integration_dispatcher/template_engine.py
```diff
@@ -99,6 +99,11 @@ def _apply_default_formatting(
             formatted_subject = f"[TEST] {subject or 'Agent Response'}"
             formatted_body = f"🧪 TEST DELIVERY\n\n{content}"
 
+        elif integration_type == IntegrationType.ZAMMAD:
+            # Zammad: No-op handler uses this; agent delivers via MCP
```
mhdawson: Does this comment still make sense "agent delivers via MCP" ?

integration-dispatcher/src/integration_dispatcher/zammad_service.py
```diff
@@ -0,0 +1,516 @@
...
+        if self.blocked_state_names and state_name in self.blocked_state_names:
+            logger.info(
+                "Zammad webhook: ticket state blocked",
+                ticket_id=ticket_id,
+                state=state_name,
+            )
+            return
+
+        tags = _ticket_tag_set(ticket)
+        if self.require_any_tags and not (tags & self.require_any_tags):
```
mhdawson: This does filtertering which I think earlier comments said it should not as the triggers will do required filtering?

test/ticket-responses-request-mgr.py
```diff
@@ -22,71 +45,93 @@
...
+    "open": "being processed by agent",
+    "pending reminder": "escalated",
+    "pending close": "escalated",
+    "closed": "closed",
+    "merged": "closed",
+}
+
+
+def _customer_password_effective(cli_password: str | None) -> str | None:
+    """Return password for customer-scoped Zammad calls, or None for legacy agent-token mode."""
```
mhdawson: What is "legacy agent-token mode" ?

mcp-servers/zammad/src/zammad_mcp/server.py
tchughesiv: fixed as discussed

integration-dispatcher/src/integration_dispatcher/integrations/zammad.py
tchughesiv: We’re trying to post as the assignee first (From = specialist), but if Zammad says no to that, we’d rather not drop the customer’s answer. With the flag on, we'd send the same public note again without From so something still gets posted, even if it shows as the API user instead of the specialist.

this is off by default. should we turn it on instead? remove entirely? i tend to lean towards keeping.

integration-dispatcher/src/integration_dispatcher/zammad_service.py
tchughesiv: fixed

evaluations/get_deepeval_metrics.py
tchughesiv: reverted

integration-dispatcher/src/integration_dispatcher/template_engine.py
tchughesiv: no, i've updated the comment

integration-dispatcher/src/integration_dispatcher/zammad_service.py
tchughesiv: removed a few code level filters that had snuck in

test/ticket-responses-request-mgr.py
tchughesiv: unnecessary logic ... simplified the method. thanks

integration-dispatcher/src/integration_dispatcher/template_engine.py
tchughesiv: to -
```python
# Customer-visible replies use ZammadIntegrationHandler (POST /ticket_articles); MCP tools are for ticket actions only.
```

helm/values-ticketing.yaml
tchughesiv: this already exists in the `main` branch -
https://github.com/rh-ai-quickstart/it-self-service-agent/blob/84cabfdc22eaa6a4828669b0b2d9c487658afa3e/helm/values.yaml#L520

with zammad channel i've just added separate default agent id. that said, these don't need to live in both helm values files so i'll remove from this one and let it live in values.yaml like the other.

agent-service/src/agent_service/session_manager.py
tchughesiv: reverted this and code related to it... will rely on the following to keep zammad "sticky" which should be sufficient. so far so good, more testing to be done.
```python
            allow_return_to_router = (
                routed_agent == self.ROUTING_AGENT_NAME
                and not self._is_zammad_integration()
            )
```
https://github.com/tchughesiv/self-service-agent-blueprint/blob/b9d6660e68ff10c896a4ee022cedc7ab6cbd780e/agent-service/src/agent_service/session_manager.py#L1006-L1009

agent-service/src/agent_service/session_manager.py
```diff
@@ -20,6 +20,24 @@
 logging.getLogger("langgraph").setLevel(logging.WARNING)
 
 
+def _default_agent_id_resolved() -> str:
+    """``DEFAULT_AGENT_ID`` — non-Zammad session bootstrap; LangGraph router identity (see ``ROUTING_AGENT_NAME``)."""
+    return os.getenv("DEFAULT_AGENT_ID", "routing-agent").strip() or "routing-agent"
+
+
+def _zammad_default_agent_id_resolved() -> str:
```
mhdawson, line 28: I'm not sure its a big deal but it still feels like the session manager has more hard code knowledge that a given communication channel uses a different default agent id. With Zammad being a hard code special case. If we were adding another communication channel that needed a third default agent then code changes would be needed to be able to support all of the communication channels at the same time which I think is where having more than one default agent is being driven from.

I'm wondering if there should be a default agent id, and then communication channels could be configured with an override. I think that would mean that the request manager would need to allow that configuration versus Zammad being handled as a special case in the session manager and pass it to the session manager.

I would probably mean that _entry_tier_agent_ids would need to read a configured set of agents or better maybe agents would need a new field that indicates if they are entry tier or not in their configuration and we should get the list that way.

tchughesiv: the only thing i'm adding is `ZAMMAD_DEFAULT_AGENT_ID` ... which is the equivalent of `DEFAULT_AGENT_ID`, but for zammad ticket routing instead. this isn't new functionality. it's matching existing functionality. 

from the `dev` branch -
https://github.com/rh-ai-quickstart/it-self-service-agent/blob/27e672cb4b0a6f284ebd2945cfdd69c4db35b55d/agent-service/src/agent_service/session_manager.py#L49

tchughesiv: Zammad is already a special case/channel because we change sessions to 1 per ticket (instead of user) and allow many per user. FWIW, all the other channels are treated the same way... so this change doesn't mean the next channel will require similar changes. That said, another ticketing platform might be able to reuse what we've already done for Zammad.

tchughesiv: would it help if we changed the name? removed "zammad" ? made this more generic/ticketing focused?

mhdawson (approved):
LGTM

PR #369, opened 2026-05-05
option to seed a knowledge base via ConfigMap

Prudhvivuda:
1. **User creates a ConfigMap** with their KB documents (each key = a file):

>     apiVersion: v1 
>     kind: ConfigMap
>     metadata:
>       name: my-kb-configmap
>     data:
>       policy.txt: |
>         Content here...
>       faq.txt: |
>         More content...

    
2. **User references it in `values.yaml`**:

>     knowledgeBases:
>       configMaps:
>         - name: my-kb-configmap   # pre-existing ConfigMap
>           kbName: my-custom-kb    # optional, defaults to ConfigMap name

    
3. **Helm renders the init job** with:
    - `EXTRA_KB_PATH=/mnt/knowledge-bases` env var
    - Each ConfigMap mounted as a subdirectory: `/mnt/knowledge-bases/my-custom-kb/`
4. **`register_assets.py`** reads `EXTRA_KB_PATH` and passes it to the KB manager.
5. **`KnowledgeBaseManager.register_knowledge_bases()`** scans both the built-in `config/knowledge_bases/` path and the extra path. Each subdirectory in the extra path becomes a KB — registered in LlamaStack exactly like the image-bundled ones.

### DIFF
--- a/agent-service/src/agent_service/knowledge/kb_manager.py
+++ b/agent-service/src/agent_service/knowledge/kb_manager.py
@@ -23,8 +23,12 @@ class KnowledgeBaseManager:
         else:
             logger.debug("Already connected to LlamaStack client")
 
-    def register_knowledge_bases(self) -> bool:
+    def register_knowledge_bases(self, extra_path: Optional[Path] = None) -> bool:
         """Register all knowledge bases by processing directories in knowledge_bases path.
+
+        Args:
+            extra_path: Optional additional directory to scan for knowledge bases
+                        (e.g. a ConfigMap mount). Each subdirectory is treated as a KB.
 
         Returns:
             bool: True if all knowledge bases were registered successfully, False otherwise.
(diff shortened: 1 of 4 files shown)


mhdawson:
@Prudhvivuda  in terms of

```text
apiVersion: v1 
kind: ConfigMap
metadata:
  name: my-kb-configmap
data:
  policy.txt: |
    Content here...
  faq.txt: |
```
can you investigate to see if there is a way to avoid having people to copy the contents of files into the config map if there is a way it can pull the content from files on the local disk?  I'm worried that escaping, indenting etc is going to be a challenge along with the extra work of cut and pasting, particularly if the content is longer. The current approach also does not support content in pdfs etc.

Prudhvivuda:
> @Prudhvivuda in terms of
> 
> ```
> apiVersion: v1 
> kind: ConfigMap
> metadata:
>   name: my-kb-configmap
> data:
>   policy.txt: |
>     Content here...
>   faq.txt: |
> ```
> 
> can you investigate to see if there is a way to avoid having people to copy the contents of files into the config map if there is a way it can pull the content from files on the local disk? I'm worried that escaping, indenting etc is going to be a challenge along with the extra work of cut and pasting, particularly if the content is longer. The current approach also does not support content in pdfs etc.

Yes, people can avoid pasting into YAML by creating the ConfigMap with kubectl create configmap … --from-file (or Kustomize/Helm from repo files). That removes escaping/indent pain. PDFs and very large files are a separate concern: ConfigMaps aren’t sized for that, and this stack currently only registers *.txt from the KB directory, so PDF support needs explicit ingestion changes plus a storage approach suited to binary/large objects (image/PVC/object store/pipeline), not a giant ConfigMap.

This reads bytes from disk and we can avoid indentation/escaping in our hand-written YAML.
```
kubectl create configmap my-kb-configmap \
  --from-file=policy.txt=./policy.txt \
  --from-file=faq.txt=./faq.txt \
  --dry-run=client -o yaml | kubectl apply -f -
```
```
kubectl create configmap my-kb-configmap \
  --from-file=./my-kb-files/ \
  --dry-run=client -o yaml | kubectl apply -f -
```

mhdawson (approved):
LGTM

PR #377, opened 2026-05-20
fix: optimize startup time

mhdawson:
It may be that the settle time is no longer needed. Setting to 0 for now. Review of the logs reports that there were no 502s or retries.

Shaves 72s off the startup time


mhdawson:
Running the short make target for the ticket flow to validate before landing.

gurvinder (approved):

PR #378, opened 2026-05-21
Update general-agent configuration and add knowledge base documents

gurvinder:
- Lowered LLM temperature from 0.7 to 0.3 for the general-agent and its LG prompt states to produce more consistent, deterministic responses
- Replaced allowed_tools: [""] with uses_mcp_tools: "No" in the LG prompt state definitions to follow the correct configuration schema
- Connected the general-support knowledge base to the general agent (knowledge_bases: ["general-support"])
- Added five new knowledge base documents covering common IT support topics: password reset, MFA authentication, BYOD network access, corporate printer use, and email phishing awareness


Makefile
```diff
@@ -1572,6 +1564,11 @@ define helm_install_common
 		--from-literal=servicenow-api-key="$${SERVICENOW_API_KEY:-}" \
 		-n $(NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
 
+	@echo "Creating additional knowledge base ConfigMaps..."
```
mhdawson: Can we add this in a section that only runs for the ticket flow. Since this is under helm_install_common I think this will be created when we install the non-ticket quickstart as well

gurvinder: moved to helm-install-ticketing target

mhdawson: I think it could probably go at the start of helm-install-ticketing:

gurvinder: sorry forgot to push, should be at the start of helm-install-ticketing now

mhdawson (approved):
LGTM

PR #380, opened 2026-05-26
feat: improve general agent use of knowledge base

mhdawson:
Based on some initial experimentation


mhdawson:
With update seems to work ok for both scout and 70b

gurvinder (approved):

PR #382, opened 2026-05-27
fix(demo-site): enforce single Zammad window to prevent cross-user session bleed

Prudhvivuda:
Summary:

- Clicking a persona in the demo site used `window.open(..., '_blank')`, creating a new Zammad tab on every sign-in. With multiple tabs open, switching users silently changed the session for all tabs, tickets appeared to be created by the wrong user.
- Even after fixing the target, the named window wasn't reloading, so the previous user's SPA state remained visible.

Changes:
- Replaced `target="_blank"` with `target="zammad-demo-window"` on all three "Open Zammad" anchor links (hero button, footer link, modal button)
- Replaced `window.open(url, '_blank', 'noopener,noreferrer')` in the persona sign-in handler with ```
window.open('', 'zammad-demo-window') + demoWin.location.href = 
ZAMMAD_BASE + '/'
``` to get a window reference and force a full reload after each login


mhdawson (approved):
LGTM

PR #384, opened 2026-05-29
Appeng 4973 test conversation through zammad UI

Alan0893:
## Description

Adds a Playwright end-to-end test that drives a full laptop-refresh conversation through the Zammad UI and validates ticket metadata (owner, group, tags) at each step by scraping the admin sidebar.

- Loads conversation messages from `evaluations/flows/ticket_laptop_refresh/conversations/success-flow-1.json`
- Logs in as Alice (customer) to create a ticket and send 4 messages
- Logs in as admin in a separate browser context to scrape and assert metadata after each agent reply
- Validates step 1: tag `agent-managed-laptop-refresh`, owner `Laptop Specialist`, group `Users`
- Validates step 4: tag `pending-manager-review` added, owner changed to `Manager One`
- Skips automatically when `ZAMMAD_URL` is not set (no impact on existing unit tests)
- Adds `make test-zammad-ui-conversation` target with `ZAMMAD_URL` validation and `$(ARGS)` passthrough

## Test

Verified passing against two live deployments:

| Deployment | Browser | Result | Time |
|---|---|---|---|
| 17B (llama-scout-17b) | Chromium | PASSED | ~86s |
| 70B (llama-3-3-70b) | Chromium | PASSED | ~95s |
| 17B | WebKit | PASSED | ~96s |
| 17B | Firefox | FAILED (SPA page load issue) | — |


evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,218 @@
...
+    admin_page: Page,
+) -> None:
+    expected_metadata: list[dict[str, Any] | None] = [
+        {
+            "tags": {"agent-managed-laptop-refresh"},
+            "owner_contains": "laptop specialist",
+            "group": "Users",
+        },
+        None,
+        None,
```
sgratch: as mentioned above, why messages/turns 2,3 are ignored?

sgratch (commented):
Few comments:
1.  For the CLI based eval tests we are currently testing the conversation for the following fields only, per each turn: `Group`, `Owner` , `State`. 

     I assume that the web based evaluation should behave the same as the  UI and therefore:
       - Since `Tags`  field are currently not tested in CLI. Maybe worth adding them to CLI as well? @mhdawson 

       - The `State` field value (new, open, closed) should be added for testing in this PR as well per each turn.

      - The fields should be verified for **each** turn. Currently IIUC only turns 1,4 are verified while turns 2,3 are ignored. The values are not changed but you still need to verify it. Is that the case?

2. Can you please rename the new folder `ui_test` to `ui_tests`? or maybe even better to `zammad_ui_tests` (since the script is specifically for zammad) ?

3. Can you please rename the `test_ui_conversation.py` file to something like: `test-ui-success-flow-1-conversation.py` to indicate which conversation we are testing?

4. Why we are hardcoding the field values in `expected_metadata` and not reading them from the evaluations/flows/ticket_laptop_refresh/conversations/success-flow-1 file, as implemented on https://github.com/rh-ai-quickstart/RAG/blob/main/evaluations/test_conversations_ui.py?  We can use the expected/actual fields for that as well...

Makefile
```diff
@@ -1430,6 +1431,20 @@ test-mock-employee-data:
 test-mock-servicenow:
 	$(call run_pytest,mock servicenow,mock-service-now)
 
+.PHONY: deps-zammad-ui-tests
+deps-zammad-ui-tests:
+	@echo "Installing Zammad UI test dependencies..."
+	cd evaluations/flows/ticket_laptop_refresh/ui_test && uv sync --group dev
+	cd evaluations/flows/ticket_laptop_refresh/ui_test && uv run playwright install chromium
+
+.PHONY: test-zammad-ui-conversation
```
mhdawson, line 1383: Out other evals targets don't have 2 separate targets (deps and run) instead they are in a single target. We should update to be consistent. 

Something like
```
.PHONY: test-zammad-ui-conversation
  test-zammad-ui-conversation:
        @if [ -z "$(ZAMMAD_URL)" ]; then \
                echo "Error: ZAMMAD_URL is required"; \
                exit 1; \
        fi
        cd evaluations/flows/ticket_laptop_refresh/ui_test && uv sync --group dev && uv run playwright install chromium && uv run python -m pyt
  test_ui_conversation.py -v $(ARGS)
```

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,218 @@
...
+    ctx.close()
+
+
+def test_laptop_refresh_conversation(
+    zammad_url: str,
+    reply_timeout: int,
+    customer_page: Page,
+    admin_page: Page,
+) -> None:
+    expected_metadata: list[dict[str, Any] | None] = [
```
mhdawson: If you rebase on the latest in dev, you'll see tht the file success-flow-1.json now has the expected metadata for each user message. 

We should use that instead of defining in here.

Makefile
```diff
@@ -1430,6 +1431,14 @@ test-mock-employee-data:
 test-mock-servicenow:
 	$(call run_pytest,mock servicenow,mock-service-now)
 
+.PHONY: test-zammad-ui-conversation
+test-zammad-ui-conversation:
+	@if [ -z "$(ZAMMAD_URL)" ]; then \
```
mhdawson: We should be able to generate the ZAMMAD_URL automatically so that if the namespace is set the user does not need to set anything manually.

Something like:

```
(evaluations) midawson@midawson-virtualbox:~/newpull/allan/it-self-service-agent$ git diff Makefile
diff --git a/Makefile b/Makefile
index f6de2b6..1e9ddc7 100644
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,8 @@ helm_demo_email_args = \
 
 # Ticketing: Zammad in-cluster URL and MCP secret wiring
 ZAMMAD_CREDENTIALS_SECRET ?= $(MAIN_CHART_NAME)-zammad-credentials
-ZAMMAD_URL ?= http://zammad-nginx.$(NAMESPACE).svc.cluster.local:8080
+_ZAMMAD_ROUTE_HOST := $(shell oc get route ssa-zammad -n $(NAMESPACE) -o jsonpath='{.spec.host}' 2>/dev/null)
+ZAMMAD_URL ?= $(if $(_ZAMMAD_ROUTE_HOST),https://$(_ZAMMAD_ROUTE_HOST),http://zammad-nginx.$(NAMESPACE).svc.cluster.local:8080)
 # Temp overlay file for cluster-specific Zammad Helm values (bootstrap image registry/tag, FQDN, OpenShift UID). Written by helm-install-ticketing.
 ZAMMAD_TICKETING_OVERLAY = /tmp/ssa-zammad-$(NAMESPACE)-overlay.yaml
 # Must match autoWizard.config in helm/values-ticketing.yaml (for UI login).
@@ -1439,11 +1440,11 @@ deps-zammad-ui-tests:
 
 .PHONY: test-zammad-ui-conversation
 test-zammad-ui-conversation: deps-zammad-ui-tests
-       @if [ -z "$(ZAMMAD_URL)" ]; then \
-               echo "Error: ZAMMAD_URL is required"; \
+       @if [ -z "$(NAMESPACE)" ] && [ -z "$${ZAMMAD_URL}" ]; then \
+               echo "Error: set ZAMMAD_URL or NAMESPACE"; \
                exit 1; \
        fi
-       cd evaluations/flows/ticket_laptop_refresh/ui_test && uv run pytest test_ui_conversation.py -v $(ARGS)
+       cd evaluations/flows/ticket_laptop_refresh/ui_test && ZAMMAD_URL=$(ZAMMAD_URL) uv run pytest test_ui_conversation.py -v $(ARGS)
 
 .PHONY: sync-evaluations
 sync-evaluations:
```

mhdawson:
Not sure if I'm running incorrectly but I'm getting assertion errors

```
                f"{expected['owner_contains']!r}, got {owner!r}"
            )
E           AssertionError: Step 1: expected owner to contain 'laptop specialist', got 'Laptop Refresh AI Agent'
E           assert 'laptop specialist' in 'laptop refresh ai agent'
E            +  where 'laptop specialist' = <built-in method lower of str object at 0x7f4cf7023930>()
E            +    where <built-in method lower of str object at 0x7f4cf7023930> = 'laptop specialist'.lower
E            +  and   'laptop refresh ai agent' = <built-in method lower of str object at 0x7f4cf6af7cb0>()
E            +    where <built-in method lower of str object at 0x7f4cf6af7cb0> = 'Laptop Refresh AI Agent'.lower

test_ui_conversation.py:91: AssertionError

```

We may need to translate from what is in the metadata in the test file to what is seen in the UI as I think they will be different as I think the existing evals use the internal Ids instead of what is visible in the UI.

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
Alan0893: > I assume that the web based evaluation should behave the same as the UI and therefore:
> 
> * Since `Tags`  field are currently not tested in CLI. Maybe worth adding them to CLI as well? @mhdawson

After the rebase, the test is now fully driven by `expected_conversation_metadata` in `success-flow-1.json`. Currently that only defines `state`, `owner`, and `group`. `tags` aren't included yet.

I agree tags should be tested. Once they're added to `expected_conversation_metadata` in the JSON, both the UI and CLI tests can pick them up from the same source of truth. Should I extend the JSON to includes `tags` as well? If not, should I drop the `tags` assertions?

mhdawson: I'd say lets add testing of the tags in a follow on PR. We can get the UI and  CLI testing to the same level and then add testing of the tags to both.

sgratch:
When trying to run it with headed mode, it failed for me with the following error:
```
           )
E           AssertionError: Step 3: expected owner to contain 'laptop' (from ID 'agent.laptop-specialist'), got 'Manager One'
E           assert 'laptop' in 'manager one'
E            +  where 'laptop' = <built-in method lower of str object at 0x7fdba87f7e70>()
E            +    where <built-in method lower of str object at 0x7fdba87f7e70> = 'laptop'.lower
E            +  and   'manager one' = <built-in method lower of str object at 0x7fdba7a57330>()
E            +    where <built-in method lower of str object at 0x7fdba7a57330> = 'Manager One'.lower

test-ui-success-flow-1-conversation.py:96: AssertionError
```

When trying to run it with headless mode , it failed for me with the following error:

```
page = <Page url='https://ssa-zammad-sgratch-it-service-agent.apps.ai-dev02.kni.syseng.devcluster.openshift.com/#ticket/zoom/193'>, count_before = 6
timeout_s = 180

    def _wait_for_agent_reply(page: Page, count_before: int, timeout_s: int) -> None:
        deadline = time.monotonic() + timeout_s
        while time.monotonic() < deadline:
            if _article_count(page) > count_before:
                time.sleep(2)
                return
            time.sleep(3)
>       raise TimeoutError(
            f"No agent reply after {timeout_s}s "
            f"(articles: before={count_before}, now={_article_count(page)})"
        )
E       TimeoutError: No agent reply after 180s (articles: before=6, now=6)

test-ui-success-flow-1-conversation.py:50: TimeoutError
-------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------
Ticket created (id=193)
Step 1: articles after creation=1, waiting for agent reply...
  Step 1 metadata: state='open', owner='Laptop Refresh AI Agent', group='Users'
Step 2: sent 'I would like to see the options', articles=3, waiting for agent reply...
  Step 2 metadata: state='open', owner='Laptop Refresh AI Agent', group='Users'
Step 3: sent '3', articles=6, waiting for agent reply...
================================================================== short test summary info ===================================================================
FAILED test-ui-success-flow-1-conversation.py::test_laptop_refresh_conversation[chromium] - TimeoutError: No agent reply after 180s (articles: before=6, now=6
```

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,242 @@
...
+def _article_count(page: Page) -> int:
+    return page.locator(".ticket-article-item").count()
+
+
+def _wait_for_agent_reply(page: Page, count_before: int, timeout_s: int) -> None:
+    deadline = time.monotonic() + timeout_s
+    while time.monotonic() < deadline:
+        page.reload()
+        page.wait_for_load_state("networkidle")
+        if _article_count(page) > count_before:
```
sgratch: I think that this check is not enough.
When running this, the created Zammad ticket constantly  includes two user turns in a row  before the agent reply appeared:

Image


Maybe you should check here if the latest article has sender_id=1 and not just that the count increased.

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,242 @@
...
+        }
+    """
+    ).strip()
+
+
+def _check_metadata_fields(
+    step: int, state: str, owner: str, group: str, expected: dict[str, Any]
+) -> None:
+    if "state" in expected:
+        assert expected["state"].lower() in state.lower(), (
```
sgratch: This will not catch all cases like empty string `""`. Please consider fixing for owner, group fields as well:

```suggestion
        assert expected["state"].lower() == state.lower(), (
```

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,242 @@
...
+
+        print(
+            f"Step {idx + 1}: sent {msg!r}, articles={count}, waiting for agent reply..."
+        )
+        _wait_for_agent_reply(customer_page, count, reply_timeout)
+
+        admin_page.reload()
+        admin_page.wait_for_load_state("networkidle")
+        time.sleep(5)
+        _assert_metadata(idx, admin_page, turns[idx]["expected_conversation_metadata"])
```
sgratch: If no "expected_conversation_metadata" exists then we should ignore the turn and skip instead of crashes.  

something like:
```
 metadata = turns[idx].get("expected_conversation_metadata")
  if metadata:
      _assert_metadata(idx, admin_page, metadata)
```

sgratch:
Please include  `evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/uv.lock` file in this PR as well

evaluations/flows/ticket_laptop_refresh/zammad_ui_tests/test-ui-success-flow-1-conversation.py
```diff
@@ -0,0 +1,277 @@
...
+    """Check whether the last article on the page was sent by an Agent (sender_id 1)."""
+    return page.evaluate("""
+        () => {
+            const items = document.querySelectorAll('.ticket-article-item');
+            if (items.length === 0) return false;
+            const lastId = items[items.length - 1].getAttribute('data-id');
+            if (!lastId) return false;
+            try {
+                const article = App.TicketArticle.find(parseInt(lastId, 10));
+                return article && article.sender_id === 1;
```
sgratch: Try to avoid magic number "1" for  `sender_id === 1` since it may  not always map to "Agent"  in a future version and it is also less readable.

So instead try implementing something like this within the JavaScript evaluate block:

```
  const agentSender = App.TicketArticleSender.findByAttribute('name', 'Agent');
  if (!agentSender) return false;
 ......
  return article && article.sender_id === agentSender.id;
```

sgratch (commented):
In current implementation you can't run the ui test more than once concurrently via 2 sessions due to  session conflict modal. 
Are we ok with that?

Alan0893:
> In current implementation you can't run the ui test more than once concurrently via 2 sessions due to session conflict modal. Are we ok with that?

You're right, that's a limitation. Since the customer and admin use separate accounts they won't conflict with each other, but two test runs against the same instance would. For now I think we can do one run at a time per instance, with parallel runs using separate namespaces.

sgratch (approved):
LGTM

mhdawson:
Landing

PR #386, opened 2026-05-29
fix(demo-site): sync Zammad session across browser windows

Prudhvivuda:
## Summary
- Add a `zammad-shell.html` wrapper that loads Zammad in an iframe and reloads when another dashboard window signs in as a different user
- Use `BroadcastChannel('zammad-demo-hub')` to coordinate session changes across separate browser windows of the same profile
- Show a session-changed banner on other dashboard windows and ping/pong before opening duplicate Zammad shells

Fixes cross-window session bleed where window 1 still showed Alice's UI after window 2 signed in as John, causing tickets to be created under the wrong user.


mhdawson (approved):
LGTM

PR #387, opened 2026-06-01
feat: add ticket-laptop-refresh evaluation targets for Scout 17b and 70b deploy configurations

gurvinder:
- Add two new Makefile targets for running the ticket-laptop-refresh evaluation with 10 generated conversations each:
  - `test-short-ticket-laptop-refresh-scout-deploy` — for clusters deployed with llama-4-scout-17b-16e-w4a16, evaluated with llama-3-3-70b-instruct-w8a8
  - `test-short-ticket-laptop-refresh-70b-deploy` — for clusters deployed with llama-3-3-70b-instruct-w8a8, evaluated with llama-3-3-70b-instruct-w8a8
- Both targets run `evaluate.py` with `-n 10` against the `ticket_laptop_refresh` and `ticket_unrelated` flows

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1448,17 +1448,11 @@ test-short-ticket-laptop-refresh:
 	uv --directory evaluations run evaluate.py --message-timeout 1800 --timeout=1800 -n 1 --flow ticket_laptop_refresh,ticket_unrelated $(VALIDATE_LAPTOP_DETAILS_FLAG) $(STRUCTURED_OUTPUT_FLAG)
 	@echo "short responses integration test for ticket-laptop-refresh flow completed successfully!"
 
-.PHONY: test-short-ticket-laptop-refresh-scout-deploy
-test-short-ticket-laptop-refresh-scout-deploy:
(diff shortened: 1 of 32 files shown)


Makefile
```diff
@@ -1448,6 +1448,18 @@ test-short-ticket-laptop-refresh:
 	uv --directory evaluations run evaluate.py --message-timeout 1800 --timeout=1800 -n 1 --flow ticket_laptop_refresh,ticket_unrelated $(VALIDATE_LAPTOP_DETAILS_FLAG) $(STRUCTURED_OUTPUT_FLAG)
 	@echo "short responses integration test for ticket-laptop-refresh flow completed successfully!"
 
+.PHONY: test-short-ticket-laptop-refresh-scout-deploy
```
mhdawson: 1) I don't think we need two of these, the same make target should work regardless of which model the quickstart was deployed with.
2) I think the name would better as something like test-long-ticket-laptop-refresh

mhdawson (approved):
LGTM

PR #389, opened 2026-06-02
Evaluation Agent Ticket Scraper

djljvv:
### **Changes**
I changed how the evaluation agent is able to receive the agent response as an input. Previously it would poll the Request Manager's /conversations API instead of scraping the ticket. This means that the created tickets for the evaluation agent weren't ever used, and this also prevented form seeing where along the process an error would occur in the case that there was some level of failure. This could also cause out of order responses from the agent response and the evaluation agent. I changed this flow such that it now takes all information from the latest relevant ticket by scraping it.
**Additions**

- Implemented helper method _zammad_get_ticket_articles() to fetch all articles from a specific Zammad ticket
- Implemented _zamma_article_ids_snapshot() to create a snapshot of current article IDs before posting a customer message
- Rewrote _poll_conversation_for_agent_reply() so it now uses the ticket itself for information and can now check and ensure proper article and sender
- Updates send_to_agent() so it now uses Zammad information rather than needing information from the RequestManager API

**Removed Code**

- _conversation_entries_first_session(): extracted conversation list form Request Manager
- _request_ids_from_conversation(): built a set of ids wanted from the request manager
- conversation_request_ids_snapshot(): made the baseline from Request Manager conversations
- RequstManagerClient import: no longer needed for polling
- _RM_CONVERSATIONS_KW constant: no more Request Manager operations

### **Tests**
With both the llama-3-3-70b and llama-scout-17b the following tests were ran:

- make test-short-ticket-laptop-refresh: PASS
- make test-short-resp-integration-request-mgr: FAIL

### DIFF
--- a/test/ticket-responses-request-mgr.py
+++ b/test/ticket-responses-request-mgr.py
@@ -2,10 +2,10 @@
 """
 CLI chat with Request Manager via Zammad ticket articles.
 
-Creates customer articles in Zammad (integration trigger → dispatcher → Request Manager)
-and polls ``GET /api/v1/conversations`` until each turn's ``agent_response`` is available.
-Conversation rows are matched by ``session_id`` ``zammad-{ticket_id}`` plus **new**
-``request_id`` per turn (snapshot taken before each article) so stale replies are not reused.
+Creates customer articles in Zammad (integration trigger → dispatcher → Zammad Articles)
+and polls Zammad for new agent-authored articles until each turn's response is available.
+Articles are matched by ``ticket_id`` and creation timestamp (snapshot taken before each
+customer article) so stale replies are not reused.
 
 Requires Zammad REST (ZAMMAD_URL, ZAMMAD_HTTP_TOKEN), Request Manager (REQUEST_MANAGER_URL),
 and ``USER_ID`` (customer email).
@@ -27,9 +27,10 @@ import os
 import sys
 import time
 from collections.abc import Awaitable, Callable
+from typing import Any
 
 import httpx
-from shared_clients import CLIChatClient, RequestManagerClient
+from shared_clients import CLIChatClient
 from shared_models.utils import normalize_zammad_rest_api_base, zammad_rest_json_headers
 
 _HTTP_TIMEOUT = 10
@@ -54,9 +55,6 @@ _ZAMMAD_JSON_CLIENT_HEADERS = {
 }
 
 DEFAULT_TICKET_TITLE = "Laptop refresh help request"
-
-# RM polling: limit 1 / latest session slice used by snapshot + agent_reply poll.
-_RM_CONVERSATIONS_KW = {"limit": 1, "offset": 0, "include_messages": True}
 
 
 def _customer_password_effective(cli_password: str | None) -> str:
@@ -185,18 +183,6 @@ def _zammad_delete_ticket(base_url: str, token: str, ticket_id: int) -> None:
     httpx.delete(url, headers=zammad_rest_json_headers(token), timeout=_HTTP_TIMEOUT)
 
 
-def _conversation_entries_first_session(data: dict) -> list:
-    """First session's ``conversation`` list from GET /conversations-style payload."""
-    sessions = data.get("sessions")
-    if not isinstance(sessions, list) or not sessions:
-        return []
-    first = sessions[0]
-    if not isinstance(first, dict):
-        return []
-    conv = first.get("conversation")
-    return conv if isinstance(conv, list) else []
-
-
 def _zammad_verify_ticket_number_logged(
     api_base: str,
     token: str,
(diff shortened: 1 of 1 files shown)


mhdawson:
@sgratch if you can review this one since it is in the evals that would be great.

test/ticket-responses-request-mgr.py
```diff
@@ -231,98 +218,129 @@ def _zammad_verify_ticket_number_logged(
...
+                    f"[zammad] warning: article id is not a positive integer, check other issues with the ticket: {article['id']}",
+                    file=sys.stderr,
+                )
+                continue
+    return out
+
+
+# completely replace the method such that we have a new version that just looks at the ticket and waits for the response to be put into the ticket
+# need to figure out how to remember the last thing on the ticket
+# get agent response from ticket: goal
```
mhdawson: These look like notes that should probably be removed.

test/ticket-responses-request-mgr.py
```diff
@@ -597,19 +609,17 @@ async def send_to_agent(msg: str) -> str:
...
             timeout_s=args.poll_timeout,
             poll_interval_s=args.poll_interval,
         )
 
     print(f"Using user ID: {user_id}")
-    print(
-        "Request Manager: poll GET /api/v1/conversations for replies (webhook-driven ingest)"
-    )
-    print("Using LangGraph state machine for conversation management")
+    print("Polling Zammad ticket articles for replies (webhook-driven ingest)")
```
sgratch: Please consider including the Zammad URL for debugging clarity instead of mentioning the webhook-driven ingest , e.g.: 
```suggestion
print(f"Polling Zammad ticket articles for replies: {zammad_base_url}")

```

test/ticket-responses-request-mgr.py
```diff
@@ -597,19 +609,17 @@ async def send_to_agent(msg: str) -> str:
...
+    print("Using Zammad ticket articles for conversation management")
```
sgratch: This seems still accurate, why removing?

```suggestion
    print("Using LangGraph state machine for conversation management") 
```

test/ticket-responses-request-mgr.py
```diff
@@ -597,19 +609,17 @@ async def send_to_agent(msg: str) -> str:
         if article_id is None:
             return "Error: could not create customer article in Zammad"
         return await _poll_conversation_for_agent_reply(
```
sgratch: Consider renaming:
```suggestion
        return await _poll_zammad_articles_for_agent_reply(
```

test/ticket-responses-request-mgr.py
```diff
@@ -231,98 +217,125 @@ def _zammad_verify_ticket_number_logged(
...
+
+"""
+update methods to return the reply from the ticket
+issue is the evaluation manager can post immediately, creating an out of order reponse
+nothing is being done with the ticket, its just kind of there
+generate ticket -> convo on ticket -> sends it to evaluation agent -> get response -> ticket is updated with the response but not used by the request manager
+
+please replace entire version of _poll_conversation_for_agent reply with a new version that gets the information from the ticket itself
+
+"""
```
sgratch: what is this docstring? it's not attached to any function, please remove

test/ticket-responses-request-mgr.py
```diff
@@ -231,98 +217,125 @@ def _zammad_verify_ticket_number_logged(
...
 
     while time.monotonic() < deadline:
         try:
-            data = await rm_client.get_conversations(
-                session_id=session_id,
-                **_RM_CONVERSATIONS_KW,
+            articles = await _zammad_get_ticket_articles(
+                zammad_base_url, zammad_token, ticket_id
             )
         except Exception as ex:
```
sgratch: `_zammad_get_ticket_articles` already catches exceptions and returns `[]`, So the `try/except` here seems redundant

test/ticket-responses-request-mgr.py
```diff
@@ -231,98 +217,125 @@ def _zammad_verify_ticket_number_logged(
...
+            article_id = entry.get("id")
+            sender_id = entry.get("sender_id")
+            body = entry.get("body")
+            if article_id is None:
+                continue
+            if int(article_id) in baseline_article_ids:
                 continue
-            rid = entry.get("request_id")
-            if rid is None or str(rid) in baseline_request_ids:
+            if sender_id != 1:
```
sgratch: `1` is hardcoded magic number. Can this` 1 `value that represents the  "Agent" differ across Zammad installations? 

Even if not, worth defining a constant like:
` ZAMMAD_AGENT_SENDER_ID = 1`

sgratch (commented):
Looks good and works ok to me.
Please check few minor comments below.

mhdawson:
Closed an reopened so that we get all CI checks to run

mhdawson:
Please rebase on the latest version of Dev. I can see it is still running such that it is using 0.0.13 instead of the current 0.0.14

test/ticket-responses-request-mgr.py
```diff
@@ -232,133 +217,118 @@ def _zammad_verify_ticket_number_logged(
...
+    base_url: str, token: str, ticket_id: int
+) -> set[int]:
+    """Returns all article ids from the zammad ticket that was created from the zammad article dictionary"""
     try:
-        data = await rm_client.get_conversations(
-            session_id=session_id,
-            **_RM_CONVERSATIONS_KW,
-        )
+        data = await _zammad_get_ticket_articles(base_url, token, ticket_id)
     except Exception:
```
sgratch: `_zammad_get_ticket_articles` already catches exceptions and returns `[]`, So the `try/except` here seems redundant  - this need to be fixed here as well

test/ticket-responses-request-mgr.py
sgratch: This doc should mention Zammad articles instead of the request manager polling.

sgratch: This still references "conversations polls" instead of "zammad polls"

test/ticket-responses-request-mgr.py
```diff
@@ -515,7 +507,7 @@ def _build_arg_parser() -> argparse.ArgumentParser:
         default=TRIGGER_POLL_INTERVAL,
         dest="poll_interval",
         help=(
-            "Seconds between GET /api/v1/conversations polls "
+            "Seconds between Zammad ticket article polls "
```
sgratch:
```suggestion
            "Seconds between Zammad ticket articles polls (GET /api/v1/ticket_articles/by_ticket/{ticket_id}) "
```

sgratch (approved):
LGTM

PR #392, opened 2026-06-03
add known-bad evals for missing close/escalate confirmation

Prudhvivuda:
- Add known-bad conversations for ticket_laptop_refresh and ticket_unrelated where the user requests ticket close or escalation but the final assistant confirmation message is missing - reproducing the production bug where ticket state changes without the user seeing a response.

- Add a deterministic CloseOrEscalationConfirmationEval metric and tighten Process Completion / Ticket Resolution prompts so evals fail when close/escalate requests are not followed by an assistant confirmation.

- Extend make check-known-bad-conversations to run known-bad checks for both ticket flows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1383,6 +1383,12 @@ check-known-bad-conversations: sync-evaluations
 	uv --directory evaluations run evaluate.py --check
 	@echo "Evaluation check completed successfully!"
 
+.PHONY: check-known-bad-ticket-laptop-refresh
+check-known-bad-ticket-laptop-refresh: sync-evaluations
+	@echo "Running evaluation check on known bad ticket laptop refresh and ticket_unrelated flows..."
+	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
+	@echo "Evaluation check completed successfully!"
+
 .PHONY: test-short-ticket-laptop-refresh
 test-short-ticket-laptop-refresh:
 	@echo "Running short responses integration test for ticket-laptop-refresh flow..."
--- a/evaluations/flows/ticket_laptop_refresh/metrics.py
+++ b/evaluations/flows/ticket_laptop_refresh/metrics.py
@@ -11,8 +11,8 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import ConversationalTestCase, TurnParams
-from helpers.close_escalation_confirmation_deterministic_eval import (
-    CloseOrEscalationConfirmationDeterministicEval,
+from helpers.user_turn_assistant_reply_deterministic_eval import (
+    UserTurnAssistantReplyDeterministicEval,
 )
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
@@ -352,7 +352,7 @@ def get_metrics(
                 "IMPORTANT: Do NOT require the agent to confirm the ticket was sent. Do NOT fail this metric because the ticket was never sent. Do NOT consider whether the laptop selection was valid or invalid.",
             ],
         ),
-        CloseOrEscalationConfirmationDeterministicEval(
+        UserTurnAssistantReplyDeterministicEval(
             threshold=1.0,
         ),
         ConversationMetadataDeterministicEval(
--- a/evaluations/flows/ticket_laptop_refresh/metrics.py
+++ b/evaluations/flows/ticket_laptop_refresh/metrics.py
@@ -11,6 +11,9 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import ConversationalTestCase, TurnParams
+from helpers.close_escalation_confirmation_deterministic_eval import (
+    CloseOrEscalationConfirmationDeterministicEval,
+)
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
 )
@@ -198,6 +201,7 @@ def get_metrics(
                 "  3. Ticket closed — the agent states the ticket was closed",
                 "PASS this metric if and ONLY if the agent explicitly confirms one of these three actions.",
                 "FAIL this metric if the conversation ends with anything other than one of these three actions — including but not limited to: creating a ticket, asking a question, giving a vague statement, or simply stopping.",
+                "FAIL this metric if the user requests ticket close or escalation and the conversation ends without an assistant confirmation message following that request — even if the ticket state may have changed in the background.",
                 "CRITICAL: The confirmation must be a statement that the action was performed, not a question asking whether to perform it.",
             ],
         ),
@@ -347,6 +351,9 @@ def get_metrics(
                 "FAIL this metric ONLY if: the ticket is explicitly sent to the manager WITHOUT the agent having asked for confirmation first, or before the user responds.",
                 "IMPORTANT: Do NOT require the agent to confirm the ticket was sent. Do NOT fail this metric because the ticket was never sent. Do NOT consider whether the laptop selection was valid or invalid.",
             ],
+        ),
+        CloseOrEscalationConfirmationDeterministicEval(
+            threshold=1.0,
         ),
         ConversationMetadataDeterministicEval(
             name="Correct conversation metadata — deterministic (predefined)",
--- a/evaluations/flows/ticket_laptop_refresh/metrics.py
+++ b/evaluations/flows/ticket_laptop_refresh/metrics.py
@@ -11,14 +11,14 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import ConversationalTestCase, TurnParams
-from helpers.user_turn_assistant_reply_deterministic_eval import (
-    UserTurnAssistantReplyDeterministicEval,
-)
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
 )
 from helpers.conversation_metadata_llm_eval import ConversationMetadataLLMEval
 from helpers.load_conversation_context import load_default_context
+from helpers.user_turn_assistant_reply_deterministic_eval import (
+    UserTurnAssistantReplyDeterministicEval,
+)
 
 # Context directory for this flow (populated at eval time by copy_flow_context).
 # Path is relative to the evaluations/ working directory.
--- a/evaluations/flows/ticket_unrelated/metrics.py
+++ b/evaluations/flows/ticket_unrelated/metrics.py
@@ -5,13 +5,13 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import TurnParams
-from helpers.user_turn_assistant_reply_deterministic_eval import (
-    UserTurnAssistantReplyDeterministicEval,
-)
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
 )
 from helpers.conversation_metadata_llm_eval import ConversationMetadataLLMEval
+from helpers.user_turn_assistant_reply_deterministic_eval import (
+    UserTurnAssistantReplyDeterministicEval,
+)
 
 
 def get_metrics(
--- a/evaluations/flows/ticket_unrelated/metrics.py
+++ b/evaluations/flows/ticket_unrelated/metrics.py
@@ -5,8 +5,8 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import TurnParams
-from helpers.close_escalation_confirmation_deterministic_eval import (
-    CloseOrEscalationConfirmationDeterministicEval,
+from helpers.user_turn_assistant_reply_deterministic_eval import (
+    UserTurnAssistantReplyDeterministicEval,
 )
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
@@ -79,7 +79,7 @@ def get_metrics(
                 "PASS if the agent responds normally even if it cannot fully resolve the user's issue.",
             ],
         ),
-        CloseOrEscalationConfirmationDeterministicEval(
+        UserTurnAssistantReplyDeterministicEval(
             threshold=1.0,
         ),
         ConversationMetadataDeterministicEval(
--- a/evaluations/flows/ticket_unrelated/metrics.py
+++ b/evaluations/flows/ticket_unrelated/metrics.py
@@ -5,6 +5,9 @@ from typing import Any, List, Optional
 from deepeval.metrics import ConversationalGEval
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.test_case import TurnParams
+from helpers.close_escalation_confirmation_deterministic_eval import (
+    CloseOrEscalationConfirmationDeterministicEval,
+)
 from helpers.conversation_metadata_deterministic_eval import (
     ConversationMetadataDeterministicEval,
 )
@@ -45,6 +48,7 @@ def get_metrics(
                 "  - The ticket has been closed (e.g. 'Your ticket has been closed')",
                 "  - The ticket has been escalated for human review (e.g. 'Your ticket has been escalated for human review')",
                 "FAIL if the conversation ends without either of these confirmations.",
+                "FAIL if the user requests ticket close or escalation and the conversation ends without an assistant confirmation message following that request — even if the ticket state may have changed in the background.",
             ],
         ),
         ConversationalGEval(
@@ -74,6 +78,9 @@ def get_metrics(
                 "  - Tool call failures or stack traces visible in the response",
                 "PASS if the agent responds normally even if it cannot fully resolve the user's issue.",
             ],
+        ),
+        CloseOrEscalationConfirmationDeterministicEval(
+            threshold=1.0,
         ),
         ConversationMetadataDeterministicEval(
             name="Correct conversation metadata — deterministic (predefined)",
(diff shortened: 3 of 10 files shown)


mhdawson:
In term sof -> Extend make check-known-bad-conversations to run known-bad checks for both ticket flows, and ignore generated known_bad_results/ eval output.

We could just run them individually before, so not sure we need to change that. I also don't understand the need to ignore generated know_bad_results/eval output ? Can you expand on what you mean by that?

EDIT
If it's just this 

```
	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
```

in the target to check known bad cases that makes sense, but still wondering that the comment "ignore generated know_bad_results/eval output" ?

mhdawson:
One more general question, was the deterministic metric needed because you could not get the prompt based ones to always catch the failure?

Prudhvivuda:
> In term sof -> Extend make check-known-bad-conversations to run known-bad checks for both ticket flows, and ignore generated known_bad_results/ eval output.
> 
> We could just run them individually before, so not sure we need to change that. I also don't understand the need to ignore generated know_bad_results/eval output ? Can you expand on what you mean by that?
> 
> EDIT If it's just this
> 
> ```
> 	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
> ```
> 
> in the target to check known bad cases that makes sense, but still wondering that the comment "ignore generated know_bad_results/eval output" ?

It's just adding the folder in the `.gitignore`. Removed it from the description to reduce the confusion.

mhdawson:
Ah adding to gitignore makes sense, thanks for clarifying.

Prudhvivuda:
> One more general question, was the deterministic metric needed because you could not get the prompt based ones to always catch the failure?

  Yes, the deterministic metric was necessary — and it's the right tool for this specific check.

  The existing GEval-based metrics (Process Completion at 0.8, Ticket Resolution at 0.9) could theoretically catch this, but there are three reasons they're not sufficient on their own:

  1. LLM judges are non-deterministic. An LLM evaluating "did the conversation end without a confirmation?" might reason differently run-to-run. For a normal good conversation you can tolerate some fuzziness. For
  a known-bad conversation you need it to always fail - a flaky known-bad test that sometimes passes undermines the entire purpose of having known-bad checks.
  2. The thresholds aren't strict enough. Both relevant GEval metrics use thresholds below 1.0 (0.8 and 0.9). An LLM judge that gives the conversation a 0.85 despite the missing message would still pass the
  metric. The deterministic metric uses threshold=1.0 with no wiggle room.
  3. This is a structural property, not a semantic one. "Does the conversation end with a user message?" requires zero LLM judgment, it's a mechanical check on turn order and string content. Deterministic code is
  strictly more reliable here than asking an LLM to notice the same thing.

  The GEval prompt additions in this PR (the "FAIL if the user requests close/escalation and the conversation ends without..." lines) are still worth keeping as they reinforce the same expectation for normal eval runs where known-bad fixtures aren't involved. But for the known-bad gate specifically, the deterministic metric is what gives you the guarantee.

Prudhvivuda:
> EDIT If it's just this
> 
> ```
> 	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
> ```
> 
> in the target to check known bad cases that makes sense

Makefile change — yes, it’s just that one line

Makefile
```diff
@@ -1440,6 +1440,7 @@ sync-evaluations:
 check-known-bad-conversations: sync-evaluations
 	@echo "Running evaluation check on known bad conversations..."
 	uv --directory evaluations run evaluate.py --check
+	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
```
sgratch, line 1389: I would suggest leave this target entry as is for the (old) conversations flow.
And add this line to a **new** make target for the tickets flow, something like:

```
.PHONY: check-known-bad-ticket-laptop-refresh
check-known-bad-ticket-laptop-refresh: sync-evaluations
	@echo "Running evaluation check on known bad ticket laptop refresh and ticket_unrelated flows..."
	uv --directory evaluations run evaluate.py --check --flow ticket_laptop_refresh,ticket_unrelated
sgratch	@echo "Evaluation check completed successfully!"
```

evaluations/helpers/close_escalation_confirmation_eval.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _assistant_confirms_action(content: str, action: str) -> bool:
+    lower = content.lower()
+    if action == "close":
+        return "closed" in lower or "ticket_closed" in lower
+    return "escalated" in lower or "ticket_escalated" in lower
+
+
+class CloseOrEscalationConfirmationEval(BaseConversationalMetric):
```
sgratch: please add the deterministic word into the metric name to indicate it's a non LLM based metric.
E.g.: `CloseOrEscalationConfirmationDeterministicEval`

evaluations/flows/ticket_laptop_refresh/metrics.py
```diff
@@ -345,6 +349,10 @@ def get_metrics(
                 "IMPORTANT: Do NOT require the agent to confirm the ticket was sent. Do NOT fail this metric because the ticket was never sent. Do NOT consider whether the laptop selection was valid or invalid.",
             ],
         ),
+        CloseOrEscalationConfirmationEval(
+            name="Close or Escalation Confirmation",
+            threshold=1.0,
+        ),
         ConversationMetadataEval(
             name="Correct conversation metadata",
```
sgratch:
```suggestion
            name="Correct conversation metadata — deterministic",
```

evaluations/flows/ticket_unrelated/metrics.py
```diff
@@ -72,6 +76,10 @@ def get_metrics(
                 "PASS if the agent responds normally even if it cannot fully resolve the user's issue.",
             ],
         ),
+        CloseOrEscalationConfirmationEval(
+            name="Close or Escalation Confirmation",
```
sgratch:
```suggestion
            name="Close or Escalation Confirmation — deterministic",
```

sgratch:
- Is there a way to reproduce this scenario of missing final assistant confirmation message for closing/escalating the ticket,  not in production mode?

-  ConversationMetadataEval  deterministic metric  for — Zammad state exists and matches expected  
    CloseOrEscalationConfirmationEval deterministic metric for —agent text exists and confirms the action
    "Ticket Resolution" llm metric for - conversation ends with closed/escalated 

     So they are a bit overlapped but the new one is a deterministic metric not identical to ConversationMetadataEval - so need it :+1:

evaluations/helpers/close_escalation_confirmation_eval.py
```diff
@@ -0,0 +1,116 @@
...
+    if any(phrase in lower for phrase in _ESCALATE_REQUEST_PHRASES):
+        return "escalate"
+    return None
+
+
+def _assistant_confirms_action(content: str, action: str) -> bool:
+    lower = content.lower()
+    if action == "close":
+        return "closed" in lower or "ticket_closed" in lower
+    return "escalated" in lower or "ticket_escalated" in lower
```
sgratch: Why not using data-driven check against a list here as well?


_CLOSE_CONFIRMATION_PHRASES = (                                                                                                    
     "closed",                                                                                                                      
      "ticket_closed",                                                                                                               
)                                                                                                                                  
 
 _ESCALATE_CONFIRMATION_PHRASES = (
      "escalated",
      "ticket_escalated",
)     
  
lower = content.lower()
 phrases = _CLOSE_CONFIRMATION_PHRASES if action == "close" else _ESCALATE_CONFIRMATION_PHRASES

return any(phrase in lower for phrase in phrases)

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Any, Optional
+
+from deepeval.metrics import BaseConversationalMetric
+from deepeval.test_case import ConversationalTestCase
+
+_CLOSE_REQUEST_PHRASES = (
+    "close the ticket",
+    "close this ticket",
+    "close my ticket",
+    "please close",
```
sgratch: It won't cover all cases. For example:
"closing the ticket" , 
"close it"....

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+_ESCALATE_REQUEST_PHRASES = (
+    "escalate",
+    "human to handle",
+    "network specialist",
+    "human review",
+    "human assistance",
```
sgratch: same here:
missing "escalating" -  for  "need escalating to."  / "I'd like escalating this to a human"
and maybe more.

Makefile
Prudhvivuda: done

evaluations/helpers/close_escalation_confirmation_eval.py
Prudhvivuda: done

Prudhvivuda:
> * Is there a way to reproduce this scenario of missing final assistant confirmation message for closing/escalating the ticket,  not in production mode?
> * ConversationMetadataEval  deterministic metric  for — Zammad state exists and matches expected
>   CloseOrEscalationConfirmationEval deterministic metric for —agent text exists and confirms the action
>   "Ticket Resolution" llm metric for - conversation ends with closed/escalated
>   So they are a bit overlapped but the new one is a deterministic metric not identical to ConversationMetadataEval - so need it 👍

Yes, make check-known-bad-ticket-laptop-refresh reproduces it offline via static fixtures. Live repro: run close-mid-flow / simple-close templates against dev with make test-short-ticket-laptop-refresh, if the bug occurs, captured conversations will end on the user’s close/escalate message with no confirmation. Agree on the metric split; the new deterministic check is specifically for the missing user-visible confirmation text.

evaluations/helpers/close_escalation_confirmation_eval.py
Prudhvivuda: Good point. Refactored to use `_CLOSE_CONFIRMATION_PHRASES` / `_ESCALATE_CONFIRMATION_PHRASES` with the same any(phrase in lower …) pattern as the request checks.

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
Prudhvivuda: Fair point. Expanded the close phrase list with "closing the ticket", "close it", etc., and added a regex fallback (_CLOSE_TICKET_PATTERN) to catch morphological variants like "could you close the ticket now" that a fixed list would miss. Same pattern for escalate.

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
Prudhvivuda: Agreed. Added "escalating to", "escalate to", "speak/talk to a human", and similar phrases, plus a richer `_ESCALATE_TICKET_PATTERN` regex (mirroring the close side) to catch variants like "need escalating to" and "I'd like escalating this to a human".

evaluations/helpers/close_escalation_confirmation_eval.py
sgratch: Please rename the file by adding` _deterministic_` accordingly, just for consistency with the other custom metric [evaluations/helpers/conversation_metadata_deterministic_eval.py](url)

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
sgratch: Well, much better but I guess we won't be able to catch all user formulations requests anyway with a deterministic metric.

For example:
 Negation, E.g.
"Please **don't** escalate this to a network specialist..." 

or a question, E.g.:
"should I escalate?"

For both, it will be marked as "escalate" even though it shouldn't.

So maybe handle this two and let's leave it as is because we can't catch all cases :)

Same for close requests.

evaluations/helpers/close_escalation_confirmation_eval.py
Prudhvivuda: okay. done

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
```diff
@@ -0,0 +1,166 @@
...
+    lower = content.lower()
+    if any(phrase in lower for phrase in _CLOSE_REQUEST_PHRASES):
+        return "close"
+    if _CLOSE_TICKET_PATTERN.search(content):
+        return "close"
+    if any(phrase in lower for phrase in _ESCALATE_REQUEST_PHRASES):
+        return "escalate"
+    if _ESCALATE_TICKET_PATTERN.search(content):
+        return "escalate"
+    return None
```
sgratch: Isn't it clearer?:
```
 if any(phrase in lower for phrase in _CLOSE_REQUEST_PHRASES) or _CLOSE_TICKET_PATTERN.search(content):
          return "close"
      if any(phrase in lower for phrase in _ESCALATE_REQUEST_PHRASES) or _ESCALATE_TICKET_PATTERN.search(content):
          return "escalate"
```

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
Prudhvivuda: ok

sgratch: As we talked today and since we can't really identify all escalation/closing user requests with a deterministic metric, worth manipulating this metric for only deterministically checking if no assistant  message followed the user request.

If we can't always identify  which request it was, then checking that the user received a specific confirmation for  _CLOSE_CONFIRMATION_PHRASES or _ESCALATE_CONFIRMATION_PHRASES seems problematic. 
WDYT? Can you make that change  ?

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
sgratch: This looks great but we have a problem with covering all cases. Please see the other comment for the escalate mode: https://github.com/rh-ai-quickstart/it-self-service-agent/pull/392#discussion_r3437570803

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
Prudhvivuda: Yes, I agree. Made the changes accordingly.

evaluations/helpers/close_escalation_confirmation_deterministic_eval.py
sgratch: I wonder about this check.

It sets the metric to validate only when close/escalate user request was detected.
If not detected (wrongly or correctly) then it will pass.

E.g.
The user asks to "Mark this as resolved"
= >The metric doesn't detect it as a close request and if there is no reply then it will PASS while should be FAIL.

it might fail with other LLM based metrics but not sure.

What do we miss if we will remove this "if" :
 ```
          if action is None:
                continue
```
check that every user turn has an assistant reply and use the `_user_requests_close_or_escalate` to log the action request as a reason for failure (close / escalate  / others)  instead of filtering by it?

Comment (Prudhvivuda) [reply evaluations/helpers/close_escalation_confirmation_deterministic_eval.py]: Got it. Removed the if action is None: continue gate. The metric now checks that every user turn is followed by an assistant reply. Close/escalate phrase matching is only used to label failures as close, escalate, or other in the reason string. This catches cases like "Mark this as resolved" with no reply that phrase matching would miss.

File: evaluations/helpers/close_escalation_confirmation_deterministic_eval.py [hunk 9]
```diff
@@ -150,21 +149,20 @@ def measure(
...
+            next_turn = turns[i + 1] if i + 1 < len(turns) else None
+            if next_turn and next_turn.role == "assistant":
                 continue
 
-            next_turn = turns[i + 1] if i + 1 < len(turns) else None
-            if not next_turn or next_turn.role != "assistant":
-                failures.append(
-                    f"User requested ticket {action} but no assistant message followed"
-                )
+            request_type = _user_requests_close_or_escalate(turn.content) or "other"
```
sgratch: The "other" is not so clear when it appears as a failure reason. 
Maybe change to "general" instead?

evaluations/helpers/close_escalation_confirmation_deterministic_eval.py
```diff
@@ -120,11 +120,10 @@ def _user_requests_close_or_escalate(content: str) -> Optional[str]:
 
 class CloseOrEscalationConfirmationDeterministicEval(BaseConversationalMetric):
```
sgratch: Metric name (and file name) is now misleading — "Close or Escalation Confirmation — deterministic" but it now checks all user turns, not just close/escalate ones.

sgratch (commented):
LGTM  👍 — what's left is one small nit  and renaming the metric after changing its behavior.

evaluations/helpers/close_escalation_confirmation_deterministic_eval.py
Prudhvivuda: ok. changed to general

evaluations/helpers/close_escalation_confirmation_deterministic_eval.py
Prudhvivuda: changed the name of the file

sgratch (approved):

Prudhvivuda:
Hi @mhdawson, can you please review and approve if it looks good to you too.

evaluations/helpers/user_turn_assistant_reply_deterministic_eval.py
```diff
@@ -0,0 +1,186 @@
...
+may miss (e.g. "Mark this as resolved").
+"""
+
+import re
+from typing import Any, Optional
+
+from deepeval.metrics import BaseConversationalMetric
+from deepeval.test_case import ConversationalTestCase
+
+_CLOSE_REQUEST_PHRASES = (
```
mhdawson: I thought we talked about just having the metric validate that we got a response from the agent versus specifically checking for close or escalate?

Prudhvivuda: Good catch, removed all close/escalate phrase matching.

sgratch: @mhdawson @Prudhvivuda The close/escalate phrase matching was only used now to label the failure reason, not to gate the check. 
I thought it was nice to keep it for that purpose of more info regarding the failure, even though it requires a few extra lines of code for the analysis. 
But if it seems redundant then sure..

mhdawson (approved):
LGTM

PR #400, opened 2026-06-10
fix: avoid knowlege base lookup failures

mhdawson:
- avoid attempted use of knowledge base in state where it is not needed. Seemed to be causing some intermittent failures
- remove conflicting info that might be causing the agent to hallucinate knowledge base tool name
- runs before were about 3/50 failures and 0/50 on run after the changes

### DIFF
--- a/agent-service/config/lg-prompts/ticket-general-lg-prompt-small.yaml
+++ b/agent-service/config/lg-prompts/ticket-general-lg-prompt-small.yaml
@@ -33,14 +33,14 @@ states:
     use_conversation_history: true
     uses_mcp_tools: "No"
     prompt: |
-      You are a helpful IT support assistant. You ONLY answer from information retrieved by the file search tool.
+      You are a helpful IT support assistant. You ONLY answer from information retrieved by the file_search tool.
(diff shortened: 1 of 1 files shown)


gurvinder (approved):

sgratch (commented):
Looks good.

Small question:
Why not fixing same knowledge_search => file_search issue for the folllowing files as  well?:
 ticket-laptop-refresh-lg-prompt-small.yaml
ticket-laptop-refresh-lg-prompt-small-scout.yaml
ticket-laptop-refresh-lg-prompt-big.yaml ?

mhdawson:
@sgratch in terms of:

Why not fixing same knowledge_search => file_search issue for the folllowing files as well?:
ticket-laptop-refresh-lg-prompt-small.yaml
ticket-laptop-refresh-lg-prompt-small-scout.yaml
ticket-laptop-refresh-lg-prompt-big.yaml ?

That makes sense, can you include that in your work on seeing how consistent longer runs are and fixing up any issues.

sgratch:
> That makes sense, can you include that in your work on seeing how consistent longer runs are and fixing up any issues

Sure

PR #405, opened 2026-06-16
Disable zammadui attachments

Prudhvivuda:
- Users no longer see attachment controls in Zammad (new ticket + reply)
- File uploads are blocked even if someone tries via the API
- Stock Zammad image and upstream Helm chart are unchanged

**How it works (two layers)**
1. Block uploads (server-side)
- Small nginx edge proxy (zammad-edge) sits in front of stock Zammad
- OpenShift Route (ssa-zammad) points to zammad-edge, not zammad-nginx directly
- Edge proxy returns 403 on POST /api/v1/upload_caches/* (Zammad’s upload API)
- Uploads fail even if the UI were visible

2. Hide attachment UI (client-side)
- Edge proxy injects inline CSS into Zammad HTML pages (/, /desktop, /app) using nginx sub_filter
- CSS hides Zammad’s attachment elements (.article-attachment, .attachmentPlaceholder, file inputs, etc.)
- Applies to new ticket and reply forms (same widget in both)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1707,7 +1707,7 @@ helm-install-ticketing: namespace helm-depend
 		true)
 	@rm -f $(ZAMMAD_TICKETING_OVERLAY)
 	@echo "Waiting for Zammad deployments to be ready..."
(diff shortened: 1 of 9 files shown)


mhdawson (approved):
LGTM

PR #406, opened 2026-06-17
Avoid deploying snow mcp server for ticket deploy

Prudhvivuda:
- Disable ServiceNow components for the ticketing deploy profile (helm-install-ticketing): mcp-self-service-agent-snow and self-service-agent-mock-servicenow are no longer deployed when using helm/values-ticketing.yaml.
- Ticket flows use Zammad MCP only (ticket-laptop-refresh, ticket-general-support), so ServiceNow is unnecessary for this profile.
- Make the snow MCP rollout wait in helm_install_common conditional, matching mock ServiceNow, so ticketing installs don’t hang waiting for a deployment that isn’t created.


mhdawson (approved):
LGTM

PR #407, opened 2026-06-17
langfuse pvc cleanup

Prudhvivuda:
- When Langfuse is enabled, the chart deploys a Redis StatefulSet named `self-service-agent-langfuse-redis`, which creates a PVC like: `data-self-service-agent-langfuse-redis-0`
- `make helm-uninstall` already had PVC cleanup for Zammad and other Langfuse components (ClickHouse, MinIO), but not for Langfuse Redis. 
- We updated the Makefile cleanup regex to include langfuse-redis.


mhdawson (approved):
LGTM

PR #410, opened 2026-06-19
fix: address frequent ticket e2e ci test failures

mhdawson:
- add wait for terminating pods. Earlier versions of oc had a bug
      which caused oc exec to sometimes pick a terminating pod
    - tweak to avoid assigning ticket to manager before confirmation as this
      was occuring frequently in my testing for other PRs


gurvinder (approved):

mhdawson:
Ran 4 times with updated code
- no failures where we see the broken pipe so I think the code to check for terminating processes is working.

Seems like the success-flow-1.json is still intermittent, with it being 50/50. Going to land anyway as it was 100% before the change to try to improve that part but there is more work to do.

PR #413, opened 2026-06-22
fix for chart rendering an empty entityOperator: block

Prudhvivuda:
**PROD mode can't install**

**Root cause:** 
- When KNATIVE_EVENTING=true, the chart creates a Strimzi Kafka CR. 
- The chart always emitted `entityOperator:`, but the inner `topicOperator` / `userOperator` sections only render when `.resources` is non-empty. With default `resources: {}` in `values.yaml`, both inner blocks were skipped, leaving Strimzi with an empty entityOperator object - which it rejects as null.

**Fix:** 
- Wrap the whole entityOperator block so it only renders when at least one operator has resources defined:

### DIFF
--- a/helm/templates/kafka-cluster.yaml
+++ b/helm/templates/kafka-cluster.yaml
@@ -26,6 +26,7 @@ spec:
       transaction.state.log.replication.factor: {{ .Values.requestManagement.kafka.config.transactionStateLogReplicationFactor }}
       auto.create.topics.enable: false
     # Resources moved to KafkaNodePool
+  {{- if or .Values.requestManagement.kafka.entityOperator.topicOperator.resources .Values.requestManagement.kafka.entityOperator.userOperator.resources }}
   entityOperator:
     {{- if .Values.requestManagement.kafka.entityOperator.topicOperator.resources }}
     topicOperator:
@@ -37,6 +38,7 @@ spec:
       resources:
         {{- toYaml .Values.requestManagement.kafka.entityOperator.userOperator.resources | nindent 8 }}
     {{- end }}
+  {{- end }}
 ---
 # KafkaNodePool for managing Kafka broker storage and resources
 apiVersion: kafka.strimzi.io/v1beta2


mhdawson:
@Prudhvivuda are you sure it is actually using kafka with only that once change ?  Or is it only part of validating that we can turn on using Kafka with the ticket deployment.

Also wondering why this is needed now, because we can run the it-self-service-agent in both Kafka and non-kafka mode ?

Prudhvivuda:
> @Prudhvivuda are you sure it is actually using kafka with only that once change ? Or is it only part of validating that we can turn on using Kafka with the ticket deployment.
> 
> Also wondering why this is needed now, because we can run the it-self-service-agent in both Kafka and non-kafka mode ?

**1. Does this single change “turn on” Kafka?**

No. Kafka / Knative eventing is already gated separately by requestManagement.knative.eventing.enabled=true. When that flag is on, the chart already renders the full production eventing stack:

- `helm/templates/kafka-cluster.yaml` — Strimzi Kafka + KafkaNodePool
- `helm/templates/knative-broker.yaml` — Knative Broker with broker.class: Kafka, wired to {kafka-name}-kafka-bootstrap...:9092
- `helm/templates/knative-triggers.yaml` — all request/agent triggers
- Mock eventing is not deployed (`mock-eventing-service-deployment.yaml` requires eventing.enabled=false)

This PR only fixes a chart rendering bug in kafka-cluster.yaml: with default entityOperator.*.resources: {}, the template emitted a bare entityOperator: key with no children. Strimzi rejects that as an invalid/null object, so helm install fails before Kafka ever comes up.

The fix omits entityOperator entirely unless at least one operator has non-empty resources — it does not add new Kafka wiring.

**2. Is this only for validating Kafka with ticket deployment?**

The failure was hit in production / Knative eventing mode (`make helm-install-prod or KNATIVE_EVENTING=true`), not because of ticketing-specific logic. `values-ticketing.yaml` does not enable eventing; default `helm-install-ticketing` still uses mock eventing (`values.yaml` has `knative.eventing.enabled: false`).

If someone runs ticketing and passes KNATIVE_EVENTING=true, they would hit the same bug — but the fix applies to any Knative eventing install, not just ticketing.

**3. Why is this needed now if we support both Kafka and non-Kafka modes?**

Both modes are supported, but they are mutually exclusive at install time — controlled by requestManagement.knative.eventing.enabled.

In the default / test / ticketing path, that flag stays false (see helm/values.yaml). The chart deploys the mock eventing service instead, and kafka-cluster.yaml is not rendered at all, so this bug never shows up.

In production / Knative path (make helm-install-prod or KNATIVE_EVENTING=true), that flag is set to true. The chart then renders the Strimzi Kafka CR, Knative Kafka broker, and triggers. That is when the empty entityOperator: block caused Strimzi to reject the install.

So non-Kafka mode was never broken. Prod mode was broken whenever someone tried the real Knative+Kafka path with default values — which is likely why this surfaced now (e.g. first prod or ticketing+eventing test on OpenShift), not because the dual-mode design changed.

mhdawson (approved):
LGTM

PR #416, opened 2026-06-29
Update make oc client from 4.11.6 to stable-4.21

Prudhvivuda:
**OC client:** Update `make oc` from hardcoded 4.11.6 to `OC_VERSION=stable-4.21` (currently 4.21.20), aligned with our OpenShift 4.21.x clusters.

**Arch:** Derive the mirror download path from `ARCH` via `OC_ARCH (linux/amd64 → x86_64, linux/arm64 → arm64)` instead of hardcoding `x86_64`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,11 @@ CONTAINER_TOOL ?= podman
 REGISTRY ?= quay.io/rh-ai-quickstart
 PYTHON_VERSION ?= 3.12
 ARCH ?= linux/amd64
+# Map ARCH to the OpenShift mirror path segment (linux/amd64 → x86_64, linux/arm64 → arm64).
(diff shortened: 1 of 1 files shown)


mhdawson:
How backward compatible are the oc versions? Just wondering if this will have any effect on using older openshift version?

Prudhvivuda:
> How backward compatible are the oc versions? Just wondering if this will have any effect on using older openshift version?

OpenShift clients are generally tolerant of server skew. I smoke-tested stable-4.17 through stable-4.21 against our 4.21.16 cluster and all could run the Makefile’s read-only oc calls (clusterversion, ingresses.config, routes). As README already documents OpenShift 4.17.0+ as requirement, should be fine for 4.17+ clusters. We can override OC_VERSION if needed for older environments.

mhdawson (approved):
LGTM

PR #417, opened 2026-06-29
feat: tweak big ticket prompt

mhdawson:
- tweak to avoid early manager assignment


mhdawson:
If anything I think this made it worse, closing.

PR #422, opened 2026-07-02
Eval/gpt oss 120b evaluation

Prudhvivuda:
**Problem**

  When using GPT-OSS-120b as the evaluation judge via DeepEval, the model occasionally wraps its `ReasonScore` output in non-standard formats instead of the expected flat {"score": ..., "reason": ...}:

  - Wrapped string: {"final": "{\"score\": 8, \"reason\": \"...\"}"}
  - Malformed key: {"final{": 0, "reason": "..."}

  Both caused `ReasonScore` pydantic validation errors, crashing the evaluation mid-run and marking conversations as unevaluated.

  **Fix**

  Added a small rescue block in `evaluations/helpers/custom_llm.py` (both generate and a_generate) that activates only when score is missing from the parsed JSON and a key starting with "final" is present. It unwraps the value into the expected structure before passing it to pydantic.

  The fix is backward compatible — models that return correct JSON `({"score": ..., "reason": ...})` hit the "score" not in data guard as False and are completely unaffected.

### DIFF
--- a/evaluations/helpers/custom_llm.py
+++ b/evaluations/helpers/custom_llm.py
@@ -125,6 +125,27 @@ class CustomLLM(DeepEvalBaseLLM):
             if schema is not None and content:
                 try:
                     data = json.loads(content)
+                    # GPT-OSS-120b sometimes wraps output in {"final": "{...}"} or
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

PR #423, opened 2026-07-08
fix: tweak ticket laptop refresh lg prompts

sgratch:
Reference: https://redhat.atlassian.net/browse/APPENG-4976

This is  a follow up fix for https://github.com/rh-ai-quickstart/it-self-service-agent/pull/415 - use correct `knowledge_search` tool name and fix typos in ticket laptop refresh lg prompts.

This should avoid eval metric failures like:
`"Correct laptop options for user location [Conversational GEva] The agent did not present all laptop models for the user's location, as the context for EMEA specifies exactly 4 laptop models, but the agent only presented 3 options, and none of them match the models in the context for EMEA."`


mhdawson (approved):
LGTM

PR #427, opened 2026-07-16
Zammad Puma Concurrency

djljvv:
### **Changes**
I changed how Zammad handles concurrent requests during ticket flows. Previously Zammad deployed with a single Puma worker, meaning all HTTP requests were processed sequentially. During a ticket flow, the agent creating articles, the dispatcher delivering webhooks, and the test script polling for status all hit Zammad simultaneously, with one worker these queue up and block each other, causing higher latency and errors like "Timeout after 180s waiting for agent article" and "assistant turn has no actual metadata". I increased Puma to 4 workers so these concurrent requests are  better handled at the same time.

I also changed how the evaluation harness determines message timeouts. Previously it hardcoded 60 seconds for all flows, which is too short for ticket flows that need to poll Zammad for up to 180 seconds from what I saw in the tests. It now auto detects the appropriate timeout based on which test script is running. I'm still not completely sure if these changes are correct nor had any meaningful impact.

  **Additions**

  - Added helpers/message_timeout.py with resolve_message_timeout() to determine the correct timeout per test script, defining 60s for chat flows, 240s for ticket flows
  - Added tests/test_message_timeout.py to verify the auto-detection logic

  **Changes**

  - Set webConcurrency: 4 in helm/zammad/values.yaml to run Puma with 4 worker processes
  - Changed --message-timeout default from hardcoded 60 to none in evaluate.py, generator.py, and run_conversations.py, deferring to resolve_message_timeout() for script-aware defaults

  **Removed**

  - Hardcoded 60s timeout default across evaluation scripts
  - Webhook trigger group filter (triggerGroupNames: "Users") which broke webhook delivery when agents moved tickets between groups as it would ignore non-sepcified ones, replaced with "triggerSkipGroupFilter: true"

### **Tests**
The tests have the same output as before I made the changes, it seems there may lie an issue in openshift_chat_client.py(lines 224-232) as this part isn't returning anything proper. I am working on checking this out however I haven't included it yet as its outside of the ticket.

### DIFF
--- a/evaluations/evaluate.py
+++ b/evaluations/evaluate.py
@@ -547,9 +547,10 @@ def _parse_arguments() -> argparse.Namespace:
     parser.add_argument(
         "--message-timeout",
         type=int,
-        default=60,
-        help="Timeout in seconds for individual message send/response operations in generator.py (default: 60). "
-        "Increase for slower agents or high concurrency scenarios.",
+        default=None,
+        help="Timeout in seconds for individual message send/response operations. "
+        "Default: auto — 60 for chat-responses-request-mgr.py; "
+        "TRIGGER_POLL_TIMEOUT+60 (240s by default) for ticket-responses-request-mgr.py.",
     )
     parser.add_argument(
         "--validate-full-laptop-details",
@@ -937,7 +938,7 @@ def run_evaluation_pipeline(
     test_script: str = "chat-responses-request-mgr.py",
     reset_conversation: bool = False,
     concurrency: int = 1,
-    message_timeout: int = 60,
+    message_timeout: Optional[int] = None,
     validate_full_laptop_details: bool = True,
     use_structured_output: bool = False,
     conversation_source: str = "generate",
@@ -961,7 +962,9 @@ def run_evaluation_pipeline(
         test_script: Name of the test script to execute
         reset_conversation: Send 'reset' message at the start of each conversation
         concurrency: Number of parallel workers for generator.py (default: 1)
-        message_timeout: Timeout for individual message send/response operations (default: 60)
+        message_timeout: Timeout for individual message send/response operations, forwarded
+            to run_conversations.py/generator.py as-is. None lets those scripts pick a
+            script-aware default (see helpers.message_timeout).
         validate_full_laptop_details: Enable validation of all 15 laptop specification fields (default: True)
         use_structured_output: Enable structured output with Pydantic schema validation (default: False)
         conversation_source: "generate" to run generator.py, "export" to run export_conversations_from_api.py (default: generate)
@@ -993,9 +996,9 @@ def run_evaluation_pipeline(
         run_conversations_args = [
             "--test-script",
             test_script,
-            "--message-timeout",
-            str(message_timeout),
         ]
+        if message_timeout is not None:
+            run_conversations_args.extend(["--message-timeout", str(message_timeout)])
         if reset_conversation:
             run_conversations_args.append("--reset-conversation")
         if flow:
@@ -1031,9 +1034,9 @@ def run_evaluation_pipeline(
             str(max_turns),
             "--test-script",
             test_script,
-            "--message-timeout",
-            str(message_timeout),
         ]
+        if message_timeout is not None:
+            generator_args.extend(["--message-timeout", str(message_timeout)])
         if concurrency > 1:
             generator_args.extend(["--concurrency", str(concurrency)])
         if reset_conversation:
--- a/evaluations/evaluate.py
+++ b/evaluations/evaluate.py
@@ -1184,14 +1184,17 @@ def main() -> int:
             effective_msg_timeout = resolve_message_timeout(
                 test_script, args.message_timeout
             )
+            MAX_PIPELINE_TIMEOUT = 2700  # 45 minutes
             min_timeout = args.num_conversations * effective_msg_timeout * 3
-            args.timeout = max(600, min_timeout)
+            args.timeout = min(MAX_PIPELINE_TIMEOUT, max(600, min_timeout))
             logger.info(
                 "Auto pipeline timeout=%ss "
-                "(num_conversations=%s × message_timeout=%ss × 3, min 600s)",
+                "(num_conversations=%s × message_timeout=%ss × 3, "
+                "floor 600s, cap %ss)",
                 args.timeout,
                 args.num_conversations,
                 effective_msg_timeout,
+                MAX_PIPELINE_TIMEOUT,
             )
 
         if args.check:
--- a/evaluations/evaluate.py
+++ b/evaluations/evaluate.py
@@ -510,9 +510,8 @@ def _parse_arguments() -> argparse.Namespace:
     parser.add_argument(
         "--timeout",
         type=int,
-        default=None,
-        help="Timeout in seconds for each script execution. "
-        "Default: auto — max(600, num_conversations × max_turns × message_timeout).",
+        default=2700,
+        help="Timeout in seconds for each script execution (default: 2700s / 45 min).",
     )
     parser.add_argument(
         "--max-turns",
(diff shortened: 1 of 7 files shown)


helm/values-ticketing.yaml
```diff
@@ -18,9 +18,9 @@ ticketingZammad:
       enabled: true
       # url: omit to use http://<mainChartReleaseName>-integration-dispatcher.<namespace>.svc.cluster.local/zammad/webhook
       mainChartReleaseName: "self-service-agent"
-      # Webhook trigger: limit to group(s) by name (bootstrap resolves to ids). Override with triggerGroupIds if you prefer numeric ids.
-      triggerGroupNames: "Users"
-      # Do not POST to the agent after manager handoff (same tag as ZAMMAD_TAG_MANAGER_REVIEW / send_to_manager_review).
+      # Remove or comment out — group filter is too brittle when agents move tickets
+      # triggerGroupNames: "Users"
+      triggerSkipGroupFilter: true
```
sgratch, line 23: What scenario is solved here? 
After this fix,  can it causes the agent to process tickets in human-managed groups?

evaluations/run_conversations.py
```diff
@@ -84,12 +86,16 @@ def _parse_arguments() -> argparse.Namespace:
         flow_ticket_title = None
         flow_include_conversation_metadata = False
 
+    from helpers.message_timeout import resolve_message_timeout
+
+    message_timeout = resolve_message_timeout(test_script, args.message_timeout)
+
```
sgratch, line 101: Why not adding here as well?:

```suggestion
 if args.message_timeout is None:
        logger.info(
            "Auto message-timeout=%ss for test script %s",
            message_timeout,
            test_script,
        )
```

evaluations/evaluate.py
```diff
@@ -1174,6 +1178,22 @@ def main() -> int:
             test_script = args.test_script or "chat-responses-request-mgr.py"
             reset_conversation = args.reset_conversation
 
+        if args.timeout is None:
+            from helpers.message_timeout import resolve_message_timeout
+
+            effective_msg_timeout = resolve_message_timeout(
+                test_script, args.message_timeout
+            )
+            min_timeout = args.num_conversations * effective_msg_timeout * 3
```
sgratch: max(600, 10 × 240 × 3) = max(600, 7200) = 7200s (2 hours)
Is that sounds as a reasonable timeout for 10 conversations?

evaluations/helpers/message_timeout.py
```diff
@@ -0,0 +1,30 @@
+"""Resolve per-message timeouts for evaluation harness subprocesses."""
+
+from __future__ import annotations
```
sgratch: is it necessary  for used Python version 3.12? I think it's redundant.

evaluations/helpers/message_timeout.py
```diff
@@ -0,0 +1,30 @@
...
+
+def ticket_harness_poll_timeout() -> float:
+    """Match test/ticket-responses-request-mgr.py TRIGGER_POLL_TIMEOUT default."""
+    return float(os.environ.get("TRIGGER_POLL_TIMEOUT", "180"))
+
+
+def default_message_timeout_for_script(test_script: str) -> int:
+    """Infer a safe per-message timeout from the harness script name."""
+    if Path(test_script).name == TICKET_RESPONSES_SCRIPT:
+        return int(ticket_harness_poll_timeout() + TICKET_HARNESS_POLL_BUFFER_SEC)
```
sgratch: why ticket_harness_poll_timeout() is not returning int type instead of float?

helm/zammad/values.yaml
```diff
@@ -2,8 +2,11 @@
 # Zammad subchart values go under the "zammad" key.
 # Override via helm/values-ticketing.yaml (when deployed as a subchart of the main chart).
 
-zammad: {}
-
+zammad:
+  zammadConfig:
+    railsserver:
+      webConcurrency: 4
```
sgratch: I checked on dev02 with 20 concurrent laptop_refresh conversations (10 parallel workers × 2 each), up to 60 existing tickets. 
With 1 worker, it once reproduced 10 "Timeout after 180s" errors in one session, but most runs completed without timeouts. 
With 4 workers, no timeouts — but not consistently better: per-request latency.

Concern:  4 Puma workers means ~4x the memory/CPU/DB-connections footprint for the pod. I suggest 2 workers as a safer starting point.

sgratch (commented):
please see few comments below.

evaluations/helpers/message_timeout.py
djljvv: I only set that to a float as I wasn't sure if fractional seconds could be used at the time I made the change, but I changed it back to int after realizing that wouldn't be the case.

helm/values-ticketing.yaml
djljvv: The main thing I'm trying to solve with this is that if the agent moves to another ticket, then the filter prevents the agent from working on the new ticket. That is what this change is meant to fix. Also it cannot process the human managed tickets, and its mostly just to help the agent manage the different tickets and allow it to do multiple things, however human managed groups are under a different tag exclusion.

evaluations/evaluate.py
sgratch: Few comments regarding the changes you added:
-  min(MAX_PIPELINE_TIMEOUT, max(600, min_timeout)) -  a clamp operation nested in the main function is unnecessarily hard to read
 --> please consider moving those calculations to a helper function. Something like resolve_pipeline_timeout() alongside resolve_message_timeout() in message_timeout.py.

- Isn't an upper constant limit simpler and more predictable than the formula. If someone runs locally without `--timeout`, a default of 2700s (45 min) is reasonable — it's what CI uses. If they need more or less, they pass `--timeout `explicitly.
The formula min(2700, max(600, n × msg_timeout × 3)) tries to be smart but:
  - The × 3 multiplier is arbitrary
  - It still gets capped at 2700 anyway
  - It adds complexity for a minor edge case

helm/values-ticketing.yaml
sgratch: Can you describe a specific example scenario where the group filter (triggerGroupNames: "Users") prevents the agent from working on a new ticket?  I still don't  understand the use case this change is solving.

Also, when a ticket's group is changed to `escalated_laptop_refresh_tickets` or `human_managed_tickets` - will the webhook be triggered and the agent will continue handle it?

Please note that the tags are not preventing since current `triggerTagsExclude` only covers `pending-manager-review`.

evaluations/helpers/message_timeout.py
sgratch: ?

djljvv: Yeah its not needed I'll remove it.

evaluations/evaluate.py
djljvv: Yeah, I did add unnecessary complexity, I removed it and I just have a more basic timeout now. Let me know if it needs more tuning or if I reduced it too much

helm/values-ticketing.yaml
djljvv: I'll fix the issue of incorrect exclusion of tags. Also for the esclated laptop refresh ticket or human managed tickets, if its handed off to human it shouldn't touch it again but can respond on moved on tickets.

evaluations/evaluate.py
sgratch: @djljvv Looks ok but few  questions to consider if we set the message timeout to a const:
- Do we want different value for chat and ticketing? 
- Do we want code checking that message-timeout < timeout? This is a tricky question since we removed the auto calculation for timeout and therefore it might happen now. Since 45 min seems a good upper limit to me then maybe just add a validation warning that the per-message timeout is useless because the
  subprocess will be killed first based on timeout value....?
- I can see the timeout value is set with defaults all over the code (e.g. set to 600 within run_check_known_bad_conversations()  method). Can you clean the code so that default value will be consistent where needed?

helm/values-ticketing.yaml
sgratch: So what will happen if a ticket was escalated to human managed tickets and the user send another question after the escalation?

djljvv: The trigger tags exculsion should cause it to be basically inaccessible to the agent by preventing the webhook from firing, so it should just be managed by a human should prevent more questions unless there is a person responding.

evaluations/evaluate.py
djljvv: - For per-message timeout, then yes as there should be another auto detection for this, however for the overall pipeline timeout 2700s works as an upper bound for both.
- Yeah I'll add a warning when message-timeout >= timeout so it's clear the subprocess would get killed before a message finishes. Won't make it a hard error though.
- I see they're all inconsistent, I'll make the changes to make it consistent.

sgratch:
Closing this PR in favor of #436 , which supersedes these changes.

PR #428, opened 2026-07-16
fix: tweak ticket-general-lg-prompt and increase long test timeout

sgratch:
Reference: https://redhat.atlassian.net/browse/APPENG-4976

Including the following fixes:
  - Increase `test-long-ticket-laptop-refresh` timeout from 30 to 45 mins to prevent premature generator timeout on slower runs.

  - For the ticket_unrelated flow, refine ESCALATE, CLOSE, and QUESTION intent classification to prevent the agent's proactive close/escalate actions that were seen on nightly CI runs and local runs — e.g., before this fix the agent sometimes closed/escalated a ticket when the user said "I'll try restarting" or "thanks for the help" without specifically asking the agent to close/escalate. Failed metric: `User-Initiated Close or Escalation`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1415,7 +1415,7 @@ test-short-ticket-laptop-refresh:
 .PHONY: test-long-ticket-laptop-refresh
 test-long-ticket-laptop-refresh:
 	@echo "Running ticket-laptop-refresh evaluation (10 conversations)..."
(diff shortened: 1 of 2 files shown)


mhdawson (approved):
LGTM

PR #435, opened 2026-08-13
fix: increase Zammad Puma workers and auto-detect eval message timeout

sgratch:
Reference: https://redhat.atlassian.net/browse/APPENG-5631

A replacement of PR #427 (original author @djljvv).

Content:
  - Increase Zammad Puma workers from 1 to 2 to handle concurrent requests during ticket flows
  - Auto-detect per-message timeout based on test script type (60s for chat, 240s for ticketing)
  - Add validation warning when message-timeout >= subprocess timeout

Changes from original PR
  - Removed webhook group filter change (triggerSkipGroupFilter) -- kept existing group-based filtering
  - Simplified message_timeout module and tests
  - Fixed trailing whitespace in helm/zammad/values.yaml

### DIFF
--- a/evaluations/evaluate.py
+++ b/evaluations/evaluate.py
@@ -46,7 +46,7 @@ logger = logging.getLogger(__name__)
 
 
 def run_script(
-    script_name: str, args: Optional[List[str]] = None, timeout: int = 600
+    script_name: str, args: Optional[List[str]] = None, timeout: int = 2700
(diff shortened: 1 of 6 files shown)


sgratch:
Closing in favor of [#436](https://github.com/rh-ai-quickstart/it-self-service-agent/pull/436) which is pushed directly to the org repo.

PR #434, opened 2026-08-04
fix: tweak chat lg-prompt to reduce laptop options hallucination

sgratch:
Reference: https://redhat.atlassian.net/browse/APPENG-5915

Port anti-hallucination fixes from ticket-mode prompt to chat-mode lg-prompt-big.yaml to avoid night CI test run failures.

The failures were mainly on following conversations:
1. `failed-flow-replication-test.json` -  on "`Nightly 20 conversation test`" runs (Jul 29 – Aug 3)
2. `generated_*` - on "`Prod deploy - Nightly 20 conversation test runs`" (Aug 2, ...)

Changes:
  - Fix typo in knowledge_search query ("availble" → "available")
  - Use specific "knowledge_search tool" instead of generic "knowledge base query tool"
  - Lower temperature from 0.3 to 0.1 (matching ticket-mode prompt)
  - Add proof requirement to verify raw knowledge_search output before presenting options
  - Add self-check to prevent using training knowledge instead of tool results
  - Add current laptop model irrelevance instruction
  - Add verbatim matching requirement for knowledge base results

PR #437, opened 2026-08-19
Reverting: fix: tweak chat lg-prompt to reduce laptop options hallucination (#434)

sgratch:
This reverts PR #434 since the anti-hallucination prompt tweaks for the chat lg-prompt did not reduce "Correct laptop options for user location" failures in nightly CI runs, and may have increased their frequency.


mhdawson (approved):
LGTM

PR #438, opened 2026-08-21
chore: update to latest architecture chart for llamastack

sgratch:
Reference: https://redhat.atlassian.net/browse/APPENG-5794

Upgrade **llama-stack Helm chart** 0.8.5 -> 0.8.9 (image 0.5.2 -> 0.6.1) and **llama-stack-client** 0.5.0 -> 0.6.1 to match.

### DIFF
--- a/agent-service/pyproject.toml
+++ b/agent-service/pyproject.toml
@@ -13,8 +13,8 @@ dependencies = [
     "cloudevents>=1.10.0",
     "httpx>=0.25.0",
     "structlog>=23.2.0",
-    # LlamaStack integration (matches server version 0.5.0)
-    "llama-stack-client==0.5.0",
+    # LlamaStack integration (matches server version 0.6.1)
+    "llama-stack-client==0.6.1",
     "requests>=2.31.0",
     "self-service-agent-shared-models",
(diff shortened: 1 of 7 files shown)


mhdawson:
1) Is it possible to update to the ogx  client as well? THe lastest llamastack architecture chart pulls in ogx instead of llama stack
2) How many runs have to done with scout 17, 70b small and 70b large prompts with both the it-self-service-agent and speeding-up-ticket-resolution quickstarts ?

mhdawson:
I also realize I should have mentioned that there is a new architecture chart for ogx which is the rename of the llamastack. RJ created it and in  addition to using the OGX rename it also uses base containers built by Red Hat.

We should be trying to use that new ogx architecture chart - https://github.com/rh-ai-quickstart/ai-architecture-charts/tree/main/ogx-ai

Having said that knowing that the last llamastack version works is a good stepping stone to using the ogx-ai chart so It might still make sense to land this if we have lots of runs confirming it works well.

mhdawson:
And in the context of my last comment we'd leave the client as llamastack until we are doing a PR that updates the architecture chart to use the OGX version instead.

